PR #26, opened 2018-02-01
Implement R8_G8_SNorm and R32_G32_B32_A32_UInt pixel formats

tgjones:
This PR implements `R8_G8_SNorm` and `R32_G32_B32_A32_UInt` pixel formats and hooks these new values up to the D3D11, Metal, Vulkan, and OpenGL backends.

I have added new tests, although I thought that with the growing number of pixel formats, it might be handy to have a single test that we can extend with new pixel formats, rather than copying-and-pasting?

I'll follow up with the BC1 and BC2 compressed formats in a separate PR.

### DIFF
--- a/src/Veldrid.Tests/TextureTests.cs
+++ b/src/Veldrid.Tests/TextureTests.cs
@@ -67,14 +67,14 @@ namespace Veldrid.Tests
             }
 
             MappedResource map = GD.Map(texture, MapMode.Read, 0);
-            ushort* mappedFloatPtr = (ushort*)map.Data;
+            ushort* mappedUShortPtr = (ushort*)map.Data;
 
             for (int y = 0; y < 1024; y++)
             {
                 for (int x = 0; x < 1024; x++)
                 {
                     ushort index = (ushort)(y * 1024 + x);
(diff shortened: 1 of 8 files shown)


src/Veldrid.Tests/TextureTests.cs
```diff
@@ -69,6 +79,37 @@ public unsafe void Update_ThenMapRead_Succeeds_R16UNorm()
...
+
+            byte[] data = Enumerable.Range(0, 8 * 8 * 2).Select(i => (byte)i).ToArray();
+
+            fixed (byte* dataPtr = data)
+            {
+                GD.UpdateTexture(texture, (IntPtr)dataPtr, 8 * 8 * sizeof(byte) * 2, 0, 0, 0, 8, 8, 1, 0, 0);
+            }
+
+            MappedResource map = GD.Map(texture, MapMode.Read, 0);
+            byte* mappedFloatPtr = (byte*)map.Data;
```
mellinoe: Should be `mappedBytePtr` (copy-paste mistake)

tgjones:
Thanks - I've fixed that, and also fixed the place where I copied it from :)

mellinoe:
Great, thanks again!

PR #27, opened 2018-02-01
Add Sdl2Window constructor that allows creation from an existing window handle

tgjones:
This PR adds an second `Sdl2Window` constructor that allows creation from an existing window handle.

This is useful for embedding into editor UIs - although AFAIK, Windows is the only platform where `SDL_CreateWindowFrom` accepts the handle of a non-window-level control, which limits its usefulness.

This is a speculative PR - I don't know whether this matches your vision for the `Sdl2Window` API. It would be really useful for me because I'd like to use the `Sdl2Window` abstraction, both for the game and also embedded in a (Windows-only-at-the-moment) editor.

My implementation could undoubtedly be improved, although because this is internal and there probably won't be a 3rd way to construct windows, I didn't want to over-engineer it.


mellinoe:
This looks like a good addition to me. I'm planning on simply using Veldrid to draw my whole editor UI (with just a single window), so I most likely won't use it, but I can definitely see the application for it. The changes to `WindowParams` are a reasonable way to handle it.

LGTM 👍

PR #28, opened 2018-02-01
Implement some additional helper state descriptions

tgjones:
This PR adds these state descriptions:

* `BlendAttachmentDescription.Disabled`
* `BlendStateDescription.SingleAdditiveBlend`
* `BlendStateDescription.SingleDisabled`
* `DepthStencilStateDescription.DepthOnlyLessEqualRead`
* `RasterizerStateDescription.CullNone`

These are only quality-of-life improvements, nothing critical. XNA has all these and I personally find them useful.

Let me know if you'd like some but not others, and I can update the PR.

### DIFF
--- a/src/Veldrid/BlendStateDescription.cs
+++ b/src/Veldrid/BlendStateDescription.cs
@@ -46,6 +46,22 @@ namespace Veldrid
         };
 
         /// <summary>
+        /// Describes a blend state in which a single color target is blended with <see cref="BlendAttachmentDescription.AdditiveBlend"/>.
+        /// </summary>
+        public static readonly BlendStateDescription SingleAdditiveBlend = new BlendStateDescription
+        {
+            AttachmentStates = new BlendAttachmentDescription[] { BlendAttachmentDescription.AdditiveBlend }
+        };
+
+        /// <summary>
+        /// Describes a blend state in which a single color target is blended with <see cref="BlendAttachmentDescription.Disabled"/>.
+        /// </summary>
+        public static readonly BlendStateDescription SingleDisabled = new BlendStateDescription
+        {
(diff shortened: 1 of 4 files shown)


src/Veldrid/BlendStateDescription.cs
```diff
@@ -45,6 +45,22 @@ public BlendStateDescription(RgbaFloat blendFactor, params BlendAttachmentDescri
...
+        /// <summary>
+        /// Describes a blend state in which a single color target is blended with <see cref="BlendAttachmentDescription.AdditiveBlend"/>.
+        /// </summary>
+        public static readonly BlendStateDescription SingleAdditiveBlend = new BlendStateDescription
+        {
+            AttachmentStates = new BlendAttachmentDescription[] { BlendAttachmentDescription.AdditiveBlend }
+        };
+
+        /// <summary>
+        /// Describes a blend state in which a single color target is blended with <see cref="BlendAttachmentDescription.Disabled"/>.
```
tgjones: I'm not sure about the naming of this one, or the awkward comment (nothing is really "blended"). Suggestions welcome...

mellinoe: I think it's fine, and makes sense. We probably wouldn't want to leave "Single" out, since the number of attachments is important. I doubt we'll have anything other than "Single" versions of these helpers, but the info is still important enough to keep in the name, I think.

The rest of the wording is good in that it is consistent with the other descriptions.

mellinoe:
Looks good to me. These aren't uncommon or "esoteric" options, so having the helpers just seems like a QOL boost. I don't really a strict "minimum usefulness" bar for adding these helpers, within reason.

Thanks for the contribution!

PR #31, opened 2018-02-02
Implement BC1 and BC2 compressed pixel formats

tgjones:
(This builds on #26, so that should be merged before this.)

This wasn't as simple as #26.

For a start, Vulkan and OpenGL have separate formats for BC1-without-alpha and BC2-with-1-bit-alpha, which I hadn't seen before.

I had to update a few things to handle the different block size for BC1 (8 bytes).

And I'm not sure that updating a BC3 with a non-origin region ever worked? In any case, I think I have fixed that, and updated it to support BC1 and BC2.

I updated the BC3 test to handle multiple permutations of texture format, size, and copy origin, which is how I found the existing bugs when copying a non-origin region. There may still be more bugs in the currently untested parts of the code.

### DIFF
--- a/src/Veldrid.Tests/TextureTests.cs
+++ b/src/Veldrid.Tests/TextureTests.cs
@@ -199,15 +199,26 @@ namespace Veldrid.Tests
             }
         }
 
-        [Fact]
-        public unsafe void Copy_BC3_Unorm()
+        [Theory]
+        [InlineData(PixelFormat.BC1_Rgb_UNorm, 8, 0, 0, 64, 64)]
+        [InlineData(PixelFormat.BC1_Rgb_UNorm, 8, 8, 4, 16, 16)]
+        [InlineData(PixelFormat.BC1_Rgba_UNorm, 8, 0, 0, 64, 64)]
+        [InlineData(PixelFormat.BC1_Rgba_UNorm, 8, 8, 4, 16, 16)]
+        [InlineData(PixelFormat.BC2_UNorm, 16, 0, 0, 64, 64)]
+        [InlineData(PixelFormat.BC2_UNorm, 16, 8, 4, 16, 16)]
+        [InlineData(PixelFormat.BC3_UNorm, 16, 0, 0, 64, 64)]
+        [InlineData(PixelFormat.BC3_UNorm, 16, 8, 4, 16, 16)]
+        public unsafe void Copy_Compressed_Texture(PixelFormat format, uint blockSizeInBytes, uint srcX, uint srcY, uint copyWidth, uint copyHeight)
         {
             Texture copySrc = RF.CreateTexture(TextureDescription.Texture2D(
-                64, 64, 1, 1, PixelFormat.BC3_UNorm, TextureUsage.Staging));
+                64, 64, 1, 1, format, TextureUsage.Staging));
             Texture copyDst = RF.CreateTexture(TextureDescription.Texture2D(
-                64, 64, 1, 1, PixelFormat.BC3_UNorm, TextureUsage.Staging));
+                copyWidth, copyHeight, 1, 1, format, TextureUsage.Staging));
 
-            uint totalDataSize = copySrc.Width * copySrc.Height;
+            const int numPixelsInBlockSide = 4;
+            const int numPixelsInBlock = 16;
+
+            uint totalDataSize = copyWidth * copyHeight / numPixelsInBlock * blockSizeInBytes;
             byte[] data = new byte[totalDataSize];
 
             for (int i = 0; i < data.Length; i++)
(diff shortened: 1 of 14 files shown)


mellinoe:
Thanks @tgjones , this is a great contribution. Overall it looks really good, and the cleanup / refactoring in particular is much appreciated.

A few things after some testing:

* Previously, `UpdateTexture` would let you pass in a width/height that was not a multiple of the block size, as long as it was targeting a 1- or 2-width mip level.
* One of the samples in Veldrid-Samples (https://github.com/mellinoe/veldrid-samples/blob/master/src/AnimatedMesh/AnimatedMesh.cs) does the above and now fails on the last two mipmap levels (1- and 2-texel width). After manually clamping the width/height to 4 (on [this line](https://github.com/mellinoe/veldrid-samples/blob/master/src/SampleBase/KtxFile.cs#L212)), the sample starts working on D3D11, Vulkan, and Metal, but OpenGL doesn't. The last two mipmap levels are all black (when viewed in RenderDoc, at least -- could be a red herring).
*  A few test cases are failing for me on Windows / Vulkan:
```
[xUnit.net 00:00:06.7293782]     Veldrid.Tests.VulkanTextureTests.Copy_Compressed_Texture(format: BC1_Rgba_UNorm, blockSizeInBytes: 8, srcX: 0, srcY: 0, copyWidth: 64, copyHeight: 64) [FAIL]
[xUnit.net 00:00:06.7306897]       Assert.Equal() Failure
[xUnit.net 00:00:06.7307555]       Expected: 128
[xUnit.net 00:00:06.7307877]       Actual:   0
[xUnit.net 00:00:06.7316255]       Stack Trace:
[xUnit.net 00:00:06.7328173]         E:\projects\veldrid\src\Veldrid.Tests\TextureTests.cs(249,0): at Veldrid.Tests.TextureTestBase`1.Copy_Compressed_Texture(PixelFormat format, UInt32 blockSizeInBytes, UInt32 srcX, UInt32 srcY, UInt32 copyWidth, UInt32 copyHeight)
[xUnit.net 00:00:07.3384006]     Veldrid.Tests.VulkanTextureTests.Copy_Compressed_Texture(format: BC1_Rgba_UNorm, blockSizeInBytes: 8, srcX: 8, srcY: 4, copyWidth: 16, copyHeight: 16) [FAIL]
[xUnit.net 00:00:07.3385570]       Assert.Equal() Failure
[xUnit.net 00:00:07.3386058]       Expected: 32
[xUnit.net 00:00:07.3386404]       Actual:   64
[xUnit.net 00:00:07.3386735]       Stack Trace:
[xUnit.net 00:00:07.3387183]         E:\projects\veldrid\src\Veldrid.Tests\TextureTests.cs(249,0): at Veldrid.Tests.TextureTestBase`1.Copy_Compressed_Texture(PixelFormat format, UInt32 blockSizeInBytes, UInt32 srcX, UInt32 srcY, UInt32 copyWidth, UInt32 copyHeight)
[xUnit.net 00:00:07.5597889]     Veldrid.Tests.VulkanTextureTests.Copy_Compressed_Texture(format: BC1_Rgb_UNorm, blockSizeInBytes: 8, srcX: 8, srcY: 4, copyWidth: 16, copyHeight: 16) [FAIL]
[xUnit.net 00:00:07.5601516]       Assert.Equal() Failure
[xUnit.net 00:00:07.5603115]       Expected: 32
[xUnit.net 00:00:07.5604233]       Actual:   64
[xUnit.net 00:00:07.5604851]       Stack Trace:
[xUnit.net 00:00:07.5605726]         E:\projects\veldrid\src\Veldrid.Tests\TextureTests.cs(249,0): at Veldrid.Tests.TextureTestBase`1.Copy_Compressed_Texture(PixelFormat format, UInt32 blockSizeInBytes, UInt32 srcX, UInt32 srcY, UInt32 copyWidth, UInt32 copyHeight)
[xUnit.net 00:00:07.9319620]     Veldrid.Tests.VulkanTextureTests.Copy_Compressed_Texture(format: BC1_Rgb_UNorm, blockSizeInBytes: 8, srcX: 0, srcY: 0, copyWidth: 64, copyHeight: 64) [FAIL]
[xUnit.net 00:00:07.9321438]       Assert.Equal() Failure
[xUnit.net 00:00:07.9321906]       Expected: 128
[xUnit.net 00:00:07.9322190]       Actual:   0
[xUnit.net 00:00:07.9322471]       Stack Trace:
[xUnit.net 00:00:07.9322963]         E:\projects\veldrid\src\Veldrid.Tests\TextureTests.cs(249,0): at Veldrid.Tests.TextureTestBase`1.Copy_Compressed_Texture(PixelFormat format, UInt32 blockSizeInBytes, UInt32 srcX, UInt32 srcY, UInt32 copyWidth, UInt32 copyHeight)
```

* Some of the tests are failing on Metal for me (a bit too long to paste but I can post it somewhere if we want to investigate it specifically).

mellinoe:
I pulled down your branch and tried to fix the issues I was seeing. At this point I've managed to get everything working (at least as far as the tests and Veldrid.Samples are concerned), except for one code path: OpenGL without the `ARB_copy_image` and `ARB_direct_state_access` extensions. This is sort of a defunct code path that is really only hit on macOS, but it's still supposed to work. The four compressed texture tests you added which copy from a non-origin region are failing. I'm sure the problem is solvable, but I'm calling it for the night. I'll plug away a bit more in the morning.

Here's my branch: ://github.com/mellinoe/Veldrid/tree/compressed-pixel-formats

tgjones:
Wow, thank you for having a look at it yourself! I just saw, one minute ago, another [commit](0852e8cc6ec7e8bf07a0f1eb6a4ce9c1f8b7a82a) which looks promising - is it all fixed now?

Next time I'll run the Metal tests myself to pick up issues like that. I don't have a device capable of running Vulkan though (thanks Apple).

mellinoe:
Yes, I believe everything is all fixed now. I'll rebase the branch now.

I just realized another problem: when we add these new enum members, we need to add them to the end, or add explicit values for everything (and preserve the existing values in the process). At the moment, we will introduce a breaking change if we redefine the existing enum values by adding a new value in the middle.

The other PR actually already introduced a breaking change like that, because it inserted values into the middle of the enum. It's no big problem -- we can easily just reorder the enum. I'll need to keep a closer eye on that kind of change in the future. And perhaps introduce some kind of automated API analysis to the build.

mellinoe:
I rebased the branch and also added https://github.com/mellinoe/veldrid/commit/cf6d142884062a72d59710ff735208c1e660e213 in order to fix the enum value ordering.

I'll take a final look at this tomorrow and run some final tests before merging.

Thanks again for the great addition!

tgjones:
> At the moment, we will introduce a breaking change if we redefine the existing enum values by adding a new value in the middle.

Ah, good point. I'll avoid that in the future.

mellinoe:
I ran some more tests and everything looked good to me. I've merged this in manually, thanks for the contribution!

tgjones:
> thanks for the contribution!

You’re welcome, thank you for making it actually work :)

PR #33, opened 2018-02-04
Fix structured buffer creation in D3D11 backend

tgjones:
I'm not sure if I'm breaking some other usage with this change.

"Extended" buffer shader resource views appear to be raw views of bytes, which means that `BufferEx.ElementCount` is interpreted as the number of bytes. So in my usage, I was finding that the shader resource view for my structured buffer was clipped at N bytes, where N is the number of structures in the structured buffer.

### DIFF
--- a/src/Veldrid/D3D11/D3D11Buffer.cs
+++ b/src/Veldrid/D3D11/D3D11Buffer.cs
@@ -56,9 +56,9 @@ namespace Veldrid.D3D11
             {
                 ShaderResourceViewDescription srvDesc = new ShaderResourceViewDescription
                 {
-                    Dimension = SharpDX.Direct3D.ShaderResourceViewDimension.ExtendedBuffer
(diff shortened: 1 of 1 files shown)


mellinoe:
It's been a while since I implemented this part, but I remember thinking this was the correct way to do it originally. I kept hitting some strange error, but it's possible I was being stupid at the time. I will test this out with the [ComputeParticles](https://github.com/mellinoe/veldrid-samples/blob/master/src/ComputeParticles/ComputeParticles.cs) demo.

mellinoe:
This seems to be correct to me, and the sample runs fine for me with this change.

On my machine, the change doesn't seem to make any difference for that sample, which uses a structured buffer to store 1024 particles. Before and after, it acts the same. The docs do clearly say that ExtendedBuffer is intended for raw buffer views, though. I'm going to assume that the fact that it's working for me is just a fluke and that your change is correct.

Issue #29, opened 2018-02-01
Calling CommandList.SetVertexBuffer with a null DeviceBuffer throws a NullReferenceException


tgjones:
I'm trying to do this:

``` csharp
commandList.SetVertexBuffer(1, null);
```

in order to clear any previously-set buffers for that slot. But both `CommandList.SetVertexBuffer` and the derived implementations of `CommandList.SetVertexBufferCore` expect the `buffer` argument to be non-null.

Is this by-design, or a bug? If it's a bug I can create a PR to fix it. If it's by-design, is there a better way to do what I'm trying to do?

I suppose that what I'm trying to do is actually unnecessary, as long as the current pipeline doesn't reference a vertex buffer in that slot (and if it did I should be setting a non-null buffer anyway), but it doesn't feel right to leave the old buffer lying around.

mellinoe:
This is by-design. You don't need to un-bind the buffers that aren't used by the current Pipeline -- they are just ignored.

I guess the proper thing to do here is
1. Document this behavior, so it's less surprising
2. Throw an exception when a null buffer is passed (perhaps referencing the behavior)

tgjones:
Okay, that makes sense. I guess it's best practice not to unbind vertex buffers.

My only thought was around garbage collection - a vertex buffer *might* (unlikely) end up still referenced by a command list, long after it was last used. But that's a very minor concern.

mellinoe:
I added an extra note in the XML doc for this method which hopefully will help explain the behavior.

PR #34, opened 2018-02-04
[WIP] Allow null shader resources in resource sets

tgjones:
(This builds on #33, so that should be merged before this one. And it should not be merged as-is, because I've only implemented this change for D3D11, and only for a subset of resource types. I wanted to get your feedback on the change before I go ahead with the other resource types and backends.)

This is different from #29 - that was about vertex buffers, and this is about shader resources (textures, structured buffers).

As far as I can tell, it *is* necessary to be able to unbind shader resources. Here's a simplified version of my use-case. I have two shaders, with these resource layouts:

```
Shader 1:
0: Texture2D
1: Texture2D

Shader 2:
0: Texture2D
1: StructuredBuffer (optional)
```

I first draw objects using Shader 1, then draw objects using Shader 2, in the same command list.

In Shader 2, the `StructuredBuffer` at slot 1 might not be present. The shader queries a value in a uniform buffer to see whether it should read from the structured buffer.

But because Veldrid doesn't allow bindable resources in a `ResourceSet` to be `null`, I can't "unbind" the `Texture2D` from Shader 1 slot 1, which causes the D3D11 debug layer to show a warning that a resource with the incorrect dimension (`Texture2D`) is bound at a slot where it expected a `StructuredBuffer`.

Do you agree with this change? If so I'll finish the implementation.

### DIFF
--- a/src/Veldrid/D3D11/D3D11Buffer.cs
+++ b/src/Veldrid/D3D11/D3D11Buffer.cs
@@ -56,9 +56,9 @@ namespace Veldrid.D3D11
             {
                 ShaderResourceViewDescription srvDesc = new ShaderResourceViewDescription
                 {
-                    Dimension = SharpDX.Direct3D.ShaderResourceViewDimension.ExtendedBuffer
+                    Dimension = SharpDX.Direct3D.ShaderResourceViewDimension.Buffer
                 };
-                srvDesc.BufferEx.ElementCount = (int)(SizeInBytes / structureByteStride);
+                srvDesc.Buffer.ElementCount = (int)(SizeInBytes / structureByteStride);
                 ShaderResourceView = new ShaderResourceView(device, _buffer, srvDesc);
             }
 
--- a/src/Veldrid/D3D11/D3D11CommandList.cs
+++ b/src/Veldrid/D3D11/D3D11CommandList.cs
@@ -347,7 +347,9 @@ namespace Veldrid.D3D11
                         BindUniformBuffer(uniformBuffer, cbBase + rbi.Slot, rbi.Stages);
                         break;
                     case ResourceKind.StructuredBufferReadOnly:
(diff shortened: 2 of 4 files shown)


tgjones:
I should mention: I know D3D11 allows null resources to be bound to the pipeline, but I don't know about the other backends. I assume they are similar.

mellinoe:
My main concern here is Vulkan. My understanding of the spec (and I just re-read it a few times), is that any resources "statically used" by a pipeline MUST be bound with valid resources of the correct type. "Static use" just means there is some instruction in the shader's call tree that accesses the resource. So even if you have some dynamic branching which would result in the resources being unused, it still needs a valid binding in place.

That said, there is apparently some use for descriptor sets (Vulkan's `ResourceSets`, essentially) to contain empty bindings, or "uninitialized descriptors". But those are only permitted when the pipeline in question NEVER uses the resource declared in that slot. My understanding is that it's only allowed to encourage you to share / re-use descriptor sets among similar, but not identical, pipelines that need mostly the same resources. So perhaps it should be possible to allow a null resource, but it wouldn't help your scenario here where the resource really is accessed sometimes.

The alternative here is to just use a "static", empty buffer that you can bind instead of a null buffer.

tgjones:
Interesting - I don't know much about Vulkan, so that's new to me.

I know D3D12 has the concept of [null descriptors](https://msdn.microsoft.com/en-us/library/windows/desktop/mt709127(v=vs.85).aspx#Null_descriptors), which boils down to calling `CreateShaderResourceView` with the resource parameter set to `null`. It sounds like Vulkan doesn't have that concept.

I've now read some of the Vulkan spec and threads about this, and I think you're right - Vulkan requires a valid resource to be bound if the shader references that descriptor at all, even in a code path that isn't hit because of dynamic branching.

It sounds like the "right" answer is to create non-zero-length dummy resources (textures, structured buffers, etc) instead of trying to use `null`. (Non-zero-length because D3D11 at least doesn't allow creation of zero-length buffers.)

So, sounds like nothing needs to change in Veldrid here.

Issue #40, opened 2018-02-14
[Feature Request] Disable not needed graphics API?


Krakean:
@mellinoe 
Is there any way to disable graphics API that you don't need?
May be some global defines like VULKAN_ENABLE / D3D11_ENABLE ?
For example, in my project I would like to support only Win64 & OSX64, so I need only D3D11 & Metal support, thus would like to disable OGL/Vulkan (so as result to not have a dependency on Vk module)

Can you add such feature (if there is none yet), please?

mellinoe:
There's no way to do this at the moment, but I can see why someone might want it. However, I have to be careful about adding a ton of little "feature flags" that you can switch on and off, because it's going to blow up the test matrix for Veldrid. As you found with the `VALIDATE_USAGE` flag, it's easy for things to fall out of sync and stop working, especially if I'm not using them myself. If this is added, I need to make sure it's done in the most manageable way, and with lots of tests.

Krakean:
@mellinoe 
>As you found with the VALIDATE_USAGE flag, it's easy for things to fall out of sync and stop working
Yes, I understand it. But flags that allows you to disable unneeded graphics API seems important to me (and may be to some others who dont use all set of gfx APIs in one project), and I cant describe it as "little" feature flags :)

mellinoe:
I describe it as "little" because each individual flag will not be very hard to add (just a few #if's and some project configuration changes), but added all together they can get very cumbersome to test. For example, it makes sense to allow every backend to be disabled (so you could get rid of D3D11 when deploying your Metal application, and vice versa), so now there's 4 extra flags, all of which can be arbitrarily combined together. Add that to `VALIDATE_USAGE`, and you've already got ~30 combinations.

Another issue is that the .NET project system really doesn't handle this kind of feature flag well. I'll need to set up a framework in the project files to streamline this.

I'd still like to do it, I just need to wrap my head around what the best approach is and how to best manage the complexity.

Krakean:
@mellinoe 
Ok, got it. Will wait for these flags :)
You can start with Vulkan ;)

mellinoe:
[...]

PR #42, opened 2018-02-14
Added GFXAPI_VULKAN_ENABLED flag

Krakean:
Allows you to cut Vulkan API support (if not defined) if you don't need it, also as addition you don't have to rely on Vk dependecy.

### DIFF
--- a/src/Veldrid.StartupUtilities/Veldrid.StartupUtilities.csproj
+++ b/src/Veldrid.StartupUtilities/Veldrid.StartupUtilities.csproj
@@ -4,6 +4,7 @@
     <PackageVersion>$(AssemblyVersion)</PackageVersion>
     <TargetFramework>netstandard2.0</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <DefineConstants>$(DefineConstants);FEATURE_VULKAN_BACKEND</DefineConstants>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Veldrid.StartupUtilities/VeldridStartup.cs
+++ b/src/Veldrid.StartupUtilities/VeldridStartup.cs
@@ -2,7 +2,6 @@
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
-using Veldrid.Vk;
 using Veldrid.Sdl2;
 using Veldrid.OpenGL;
 
@@ -98,7 +97,11 @@ namespace Veldrid.StartupUtilities
                 case GraphicsBackend.Direct3D11:
                     return CreateDefaultD3D11GraphicsDevice(options, window);
                 case GraphicsBackend.Vulkan:
+#if FEATURE_VULKAN_BACKEND
                     return CreateVulkanGraphicsDevice(options, window);
+#else
+                    throw new VeldridException("Vulkan support has not been included in this configuration of Veldrid");
+#endif
                 case GraphicsBackend.OpenGL:
                     return CreateDefaultOpenGLGraphicsDevice(options, window);
                 case GraphicsBackend.Metal:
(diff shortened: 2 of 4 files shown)


feliwir:
Some thoughts:

- I think Vulkan should be enabled by default. By looking at the code it looks like the opposite is the case.
- It might be smarter to just make Vulkan not an available backend when being disabled, instead of removing every occurence of Vulkan. Having to disable much less code and requiring less changes
- I don't like Vulkan being the only backend that can be disabled. It makes it look worse, eventhough it's arguably the most useful backend 😄

mellinoe:
And yes, to echo @feliwir's sentiment: this should be a totally-optional source-compile-only change. By default, everything should still be included. I'll also want to put this into a separate branch until we have everything sorted out and all of the backends are included. You mentioned over in #40 that you haven't made the project file changes yet, which is fine. I think if you take into account the above feedback, you should not be changing very many project files.

Krakean:
@feliwir 
>I think Vulkan should be enabled by default. By looking at the code it looks like the opposite is the case.

@mellinoe 
>I don't want to proliferate these #if's all throughout every project. For example, there's really no reason to allow NeoDemo to use this flag.

Thing is I just didnt uploaded .csproj's where this defined was added by default due to reasons I described here: https://github.com/mellinoe/veldrid/issues/40#issuecomment-365733104
So, for end user actually nothing should changed (if he is using Veldrid only, without NeoDemo for example)

@feliwir 
>It might be smarter to just make Vulkan not an available backend when being disabled, instead of removing every occurence of Vulkan.

@mellinoe 
>I'm not convinced we should remove otherwise-harmless things from the public API

Agree, thought about it too. But since @mellinoe initially do not told me how he see it, I decided to be adventurous :)


@feliwir 
>I don't like Vulkan being the only backend that can be disabled
Not really, its just a start in case if we came to united solution about how such thing should be implemented at all. I have same plans for Metal/OpenGL.

mellinoe:
> Thing is I just didnt uploaded .csproj's where this defined was added by default due to reasons I described here: #40 (comment)
So, for end user actually nothing should changed (if he is using Veldrid only, without NeoDemo for example)

I'm not really sure what you mean. Currently, the code removes everything related to Vulkan if `GFXAPI_VULKAN_ENABLED` isn't set in the project file, which it isn't. Or do you mean you just forgot to add the projects to your commit?

Krakean:
@mellinoe 
>Currently, the code removes everything related to Vulkan if GFXAPI_VULKAN_ENABLED isn't set in the project file, which it isn't

Yes, I didnt add this define in .csproj file by default in this commit, due to necessity to add this define to other 3 projects (NeoDemo, StartupUtilities, ImGui) besides Veldrid, which I was thought not a very well solution and thought that you can came up with better idea (for example having like common.props with defines that common for a few projects)

mellinoe:
@Krakean If you just add the define to Veldrid, then you shouldn't need to touch any of the other projects.

> P.S. May be you have Skype or Telegram or something like this for instant messaging? It could save us some time, I suppose :)

I've been thinking of adding a gitter channel, maybe it would help for things like this.

Krakean:
@mellinoe 
Yes, please add gitter.

Krakean:
@mellinoe 
Remade commit. Now I suppose it much better & cleaner. And still works like a charm :)

Now only Veldrid & Veldrid.StartupUtilities are depend on GFXAPI_VULKAN_ENABLED.

src/Veldrid.StartupUtilities/VeldridStartup.cs
```diff
@@ -97,8 +96,10 @@ public static GraphicsDevice CreateGraphicsDevice(Sdl2Window window, GraphicsBac
             {
                 case GraphicsBackend.Direct3D11:
                     return CreateDefaultD3D11GraphicsDevice(options, window);
+#if GFXAPI_VULKAN_ENABLED
```
mellinoe: This should throw a VeldridException with some message like "Vulkan support has not been included in this configuration of Veldrid." ATM it will go through the `default` case below with a generic message.

Krakean:
@mellinoe 
Fixed, check.

src/Veldrid.StartupUtilities/Veldrid.StartupUtilities.csproj
```diff
@@ -16,4 +16,12 @@
     <PackageTags>Core Standard Game SDL2 Window Input</PackageTags>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <DefineConstants>TRACE;DEBUG;GFXAPI_VULKAN_ENABLED;NETSTANDARD2_0</DefineConstants>
```
mellinoe: These shouldn't be defined here. Just add this line tot he first `PropertyGroup` of the file:

`<DefineConstants>$(DefineConstants);GFXAPI_VULKAN_ENABLED</DefineConstants>

mellinoe:
Let's rename this define to `FEATURE_VULKAN_BACKEND` -- that's a more common pattern

mellinoe:
@Krakean I added a gitter channel: https://gitter.im/veldrid, feel free to stop in if you have questions.

mellinoe:
@Krakean This version looks okay so far. I've pushed it into the [optional-backends](https://github.com/mellinoe/veldrid/tree/optional-backends) branch. If we fill out the rest of the configuration options then we can get this merged into master.

PR #43, opened 2018-02-15
Added defines to exclude unneeded backends

Krakean:
Added defines FEATURE_D3D11_BACKEND, FEATURE_METAL_BACKEND, FEATURE_OPENGL_BACKEND, besides FEATURE_VULKAN_BACKEND. Allows you to individually disable backends that you don't need.


mellinoe:
@Krakean Thanks for the contribution. I'll merge this into the optional-backends branch, and will most likely add an extra commit into it before merging back into master. As I mentioned in gitter, we should probably change up the project file slightly so that it's possible to build Veldrid.csproj itself (and not a custom project file) without one or more backends.

PR #44, opened 2018-02-16
Fixed possible ambiguous reference

Krakean:


mellinoe:
This is from the new SharpDX beta, right? I saw this as well. I'd like to see if SharpDX will consider renaming the new type so that it doesn't conflict with `System.Diagnostics.Debug` before merging this. I think a lot of people will need to change their code to cope with that addition, which seems like something they might want to avoid if possible.

Krakean:
>This is from the new SharpDX beta, right?

Yep.

mellinoe:
Based on some discussion over in the SharpDX repo (https://github.com/sharpdx/SharpDX/pull/972), I think this will be changed in SharpDX and won't be necessary here. I'll leave this open until that is confirmed, though.

erictuvesson:
This was updated in PR [sharpdx/SharpDX#1021](https://github.com/sharpdx/SharpDX/pull/1021)

PR #49, opened 2018-03-04
Fix the SwapchainDescription.SyncToVerticalBlank value being ignored

SomeAnon42:
…when creating a Swapchain. Seems like a regression from 4.0.0.


mellinoe:
The fix looks good. Thanks for catching this. I'm going to:

* Merge this
* Add a test for Swapchain creation
* Tag a 4.1.1 version
* Upload to nuget.org

PR #54, opened 2018-03-06
Very minor code cleanup/refactor.

Krakean:
a) Removed unused members.
b) Added "I" to interfaces.
c) Compile StagingMemoryPool.cs only if OpenGL enabled, since its used only for this backend.

### DIFF
--- a/src/Veldrid.Utilities/ObjParser.cs
+++ b/src/Veldrid.Utilities/ObjParser.cs
@@ -12,7 +12,6 @@ namespace Veldrid.Utilities
     /// </summary>
     public class ObjParser
     {
-        private static readonly string[] s_newline = new string[] { "\n" };
         private static readonly char[] s_whitespaceChars = new char[] { ' ' };
         private static readonly char[] s_slashChar = new char[] { '/' };
 
@@ -561,7 +560,7 @@ namespace Veldrid.Utilities
     /// <summary>
     /// A standalone <see cref="MeshData"/> created from information from an <see cref="ObjFile"/>.
     /// </summary>
-    public class ConstructedMeshInfo : MeshData
+    public class ConstructedMeshInfo : IMeshData
     {
         /// <summary>
         /// The vertices of the mesh.
@@ -569,14 +568,14 @@ namespace Veldrid.Utilities
         public VertexPositionNormalTexture[] Vertices { get; }
 
         /// <summary>
-        /// The indices of the mesh.
-        /// </summary>
-        public ushort[] Indices { get; }
-
-        /// <summary>
         /// The name of the <see cref="MaterialDefinition"/> associated with this mesh.
         /// </summary>
         public string MaterialName { get; }
+
+        /// <summary>
+        /// The indices of the mesh.
+        /// </summary>
+        private ushort[] Indices { get; }
 
         /// <summary>
         /// Constructs a new <see cref="ConstructedMeshInfo"/>.
--- a/src/Veldrid.Utilities/SimpleMeshDataProvider.cs
+++ b/src/Veldrid.Utilities/SimpleMeshDataProvider.cs
@@ -4,11 +4,12 @@ using System.Numerics;
 
 namespace Veldrid.Utilities
 {
-    public class SimpleMeshDataProvider : MeshData
+    public class SimpleMeshDataProvider : IMeshData
     {
         public VertexPositionNormalTexture[] Vertices { get; }
-        public ushort[] Indices { get; }
         public string MaterialName { get; }
+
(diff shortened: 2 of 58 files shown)


mellinoe:
Thanks for the cleanup @Krakean .

Unfortunately, you'll need to revert all of the interface renames. Because:

1. Many of them are breaking changes. You can't rename a type and maintain compatibility with previous versions. I won't be making any breaking changes for any of the upcoming releases. Some of the interfaces are internal, but I still don't want to rename any.
2. I don't believe that prefacing interfaces with "I" is actually a good idea, personally. That's why these interfaces are named like they are.

I'll take a closer look when those are undone, since there's quite a lot of code changed because of that.

src/Veldrid/Veldrid.csproj
```diff
@@ -33,6 +33,7 @@
     <Compile Remove="D3D11/**/*" Condition="'$(ExcludeD3D11)' == 'true'" />
     <Compile Remove="Vk/**/*" Condition="'$(ExcludeVulkan)' == 'true'" />
     <Compile Remove="MTL/**/*" Condition="'$(ExcludeMetal)' == 'true'" />
+    <Compile Remove="StagingMemoryPool.cs" Condition="'$(ExcludeOpenGL)' == 'true'" />
```
mellinoe: It would be better to just move the type into the OpenGL folder and change its namespace.

src/Veldrid/Vk/StackList.cs
```diff
@@ -13,7 +13,6 @@ internal unsafe struct StackList<T> where T : struct
         public const int CapacityInBytes = 256;
         private static readonly int s_sizeofT = Unsafe.SizeOf<T>();
 
-        private fixed byte _storage[CapacityInBytes];
```
mellinoe: Revert this: it completely breaks this type. It looks unused, but it's not.

src/Veldrid/Veldrid.csproj
Krakean: Corrected.

src/Veldrid/Vk/StackList.cs
Krakean: Fixed

Krakean:
>Many of them are breaking changes. 

Which exactly I should revert? Remove "I" prefix from interfaces? 
Though, its a standart to name them like this (to differentiate intf types from impl), just a few links:
https://msdn.microsoft.com/en-us/library/8bc1fexb(v=vs.71).aspx
https://stackoverflow.com/questions/681700/interface-naming-convention
http://www.dofactory.com/reference/csharp-coding-standards

>I won't be making any breaking changes for any of the upcoming releases.

NeoDemo was compiled fine, just because these changes mostly internal to Veldrid.

mellinoe:
Yes, revert any style changes that I've not requested, like the I-prefix addition. I'm familiar with the arguments for it, I just don't agree with them and have made the decision not to follow that convention. It's hard to look through the rest of the changes because 90% of the PR is just fallout from renaming interfaces. I think there will only be a handful of changes after that.

> NeoDemo was compiled fine, just because these changes mostly internal to Veldrid.

That's because it is compiling against Veldrid directly. Renaming a type is a breaking change, so it cannot be done.

src/Veldrid.Utilities/ObjParser.cs
```diff
@@ -561,22 +560,22 @@ public Face(FaceVertex v0, FaceVertex v1, FaceVertex v2, int smoothingGroup = -1
...
+    public class ConstructedMeshInfo : IMeshData
     {
         /// <summary>
         /// The vertices of the mesh.
         /// </summary>
         public VertexPositionNormalTexture[] Vertices { get; }
 
         /// <summary>
-        /// The indices of the mesh.
+        /// The name of the <see cref="MaterialDefinition"/> associated with this mesh.
```
mellinoe: These properties don't need to be re-ordered.

src/Veldrid.Utilities/SimpleMeshDataProvider.cs
```diff
@@ -4,12 +4,13 @@
...
 namespace Veldrid.Utilities
 {
-    public class SimpleMeshDataProvider : MeshData
+    public class SimpleMeshDataProvider : IMeshData
     {
         public VertexPositionNormalTexture[] Vertices { get; }
-        public ushort[] Indices { get; }
         public string MaterialName { get; }
 
+        private ushort[] Indices { get; }
```
mellinoe: Ditto: these properties don't need to be re-ordered.

mellinoe:
Closing since there's been no activity.

PR #60, opened 2018-04-08
Add configurable option to bind vertex data buffers after other vertex-stage uniform and structured buffers in Metal backend

tgjones:
Matches the behaviour change in https://github.com/mellinoe/ShaderGen/pull/36.

`BindMetalVertexBuffersAfterOtherBuffers` is obviously just a placeholder name - do you have suggestions for a better name?

### DIFF
--- a/src/Veldrid/GraphicsPipelineDescription.cs
+++ b/src/Veldrid/GraphicsPipelineDescription.cs
@@ -33,6 +33,10 @@ namespace Veldrid
         /// </summary>
         public ResourceLayout[] ResourceLayouts;
         /// <summary>
+        /// Specifies which model the rendering backend should use for binding resources.
+        /// </summary>
+        public ResourceBindingModel ResourceBindingModel;
+        /// <summary>
         /// A description of the output attachments used by the <see cref="Pipeline"/>.
         /// </summary>
         public OutputDescription Outputs;
@@ -67,6 +71,7 @@ namespace Veldrid
             PrimitiveTopology = primitiveTopology;
             ShaderSet = shaderSet;
             ResourceLayouts = resourceLayouts;
+            ResourceBindingModel = ResourceBindingModel.Default;
             Outputs = outputs;
         }
 
@@ -100,6 +105,7 @@ namespace Veldrid
             PrimitiveTopology = primitiveTopology;
             ShaderSet = shaderSet;
             ResourceLayouts = new[] { resourceLayout };
+            ResourceBindingModel = ResourceBindingModel.Default;
             Outputs = outputs;
         }
 
@@ -116,6 +122,7 @@ namespace Veldrid
                 && PrimitiveTopology == other.PrimitiveTopology
                 && ShaderSet.Equals(other.ShaderSet)
                 && Util.ArrayEquals(ResourceLayouts, other.ResourceLayouts)
+                && ResourceBindingModel.Equals(other.ResourceBindingModel)
                 && Outputs.Equals(other.Outputs);
         }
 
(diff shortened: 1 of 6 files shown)


mellinoe:
This looks roughly like what I was thinking. I need to think over how this will interact with some other stuff cleanly. I think this option will need to be configurable on a per-Pipeline basis. Otherwise, it will be too difficult to provide re-usable components. For example, Veldrid.ImGui is supposed to be usable from any GraphicsDevice, regardless of which binding model it is using. So, it should be able to specify the legacy binding model when it creates its Pipeline.

As for naming, I'm thinking something like:

```C#
public enum ResourceBindingModel
{
    Default = 0,
    Improved = 1,
}
```

If a Pipeline doesn't specify the binding model, then it defaults to the GraphicsDevice's default, as you've done.

Ultimately, this whole enum (and support for the legacy binding model) should be removed at the next major version bump, but that's something I wasn't planning on doing for a long time. This will definitely be a positive change, but the transition will be a bit more painful than I was hoping 😞.

mellinoe:
Sorry I wasn't as clear as I could have been -- it's a bit late here 😴 .

I actually had thought that the GraphicsDevice option would remain as the "default" if it wasn't overridden in the pipeline description. This would mean that the GraphicsPipelineDescription field would actually be a `ResourceBindingModel?` (nullable). If unspecified/null, then the GraphicsDevice's default is chosen. There should be another constructor for GraphicsPipelineDescription which takes a ResourceBindingModel (non-nullable).

I'm still mulling this over, but that was my original idea. I'll think about this some more tomorrow and see if what I wrote above makes sense.

Aside: ComputePipelineDescription will eventually need this option too, but only when I get around to implementing the improved OpenGL binding model. For Metal it doesn't matter.

tgjones:
To help with the mulling, I've changed it to (hopefully) match what you've described.

Krakean:
@mellinoe 
>but that's something I wasn't planning on doing for a long time. 

Well, if you will make major version bump, please clean public API from unused params/functions (which I partialy tried to do in some of my PR that wasn't accepted), because their presence is confusing.

src/Veldrid/GraphicsPipelineDescription.cs
```diff
@@ -100,6 +106,44 @@ public struct GraphicsPipelineDescription : IEquatable<GraphicsPipelineDescripti
...
+        /// the value specified in <see cref="GraphicsDeviceOptions"/>.</param>
+        /// <param name="outputs">A description of the output attachments used by the <see cref="Pipeline"/>.</param>
+        public GraphicsPipelineDescription(
+            BlendStateDescription blendState,
+            DepthStencilStateDescription depthStencilStateDescription,
+            RasterizerStateDescription rasterizerState,
+            PrimitiveTopology primitiveTopology,
+            ShaderSetDescription shaderSet,
+            ResourceLayout[] resourceLayouts,
+            ResourceBindingModel resourceBindingModel,
```
mellinoe: I'll merge as-is but I'm going to swap the last two parameters so they're in the same order as the previous overload.

mellinoe:
@tgjones Looks good to me. I'll make a small followup commit (to swap the parameter order).

I'm going to continue thinking about this before 4.2.0 goes stable, but I think this is a decent compromise to make a much-needed improvement while maintaining compatibility.

tgjones:
Thank you! Please could you also merge https://github.com/mellinoe/ShaderGen/pull/36 (when you have time)?

mellinoe:
@tgjones Merged that as well.

PR #62, opened 2018-04-09
Add missing depth buffer formats to the OpenGL backend

feliwir:
Those formats were supported by all other backends, but still missing for the OpenGL backend.

### DIFF
--- a/src/Veldrid.StartupUtilities/VeldridStartup.cs
+++ b/src/Veldrid.StartupUtilities/VeldridStartup.cs
@@ -232,7 +232,7 @@ namespace Veldrid.StartupUtilities
             if (backend == GraphicsBackend.OpenGL)
             {
                 Sdl2Native.SDL_GL_SetAttribute(SDL_GLAttribute.ContextProfileMask, (int)SDL_GLProfile.Core);
-            Sdl2Native.SDL_GL_SetAttribute(SDL_GLAttribute.ContextMajorVersion, 3);
+                Sdl2Native.SDL_GL_SetAttribute(SDL_GLAttribute.ContextMajorVersion, 3);
                 Sdl2Native.SDL_GL_SetAttribute(SDL_GLAttribute.ContextMinorVersion, 0);
             }
             else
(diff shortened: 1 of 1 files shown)


src/Veldrid.StartupUtilities/VeldridStartup.cs
```diff
@@ -243,21 +243,31 @@ public static unsafe GraphicsDevice CreateDefaultOpenGLGraphicsDevice(GraphicsDe
...
+                    case PixelFormat.D32_Float_S8_UInt:
+                        depthBits = 32;
+                        stencilBits = 8;
+                        break;
                     default:
                         throw new VeldridException("Invalid depth format: " + options.SwapchainDepthFormat.Value);
                 }
             }
 
+            Sdl2Native.SDL_GL_SetAttribute(SDL_GLAttribute.StencilSize, stencilBits);
```
mellinoe: double nit: Can you put this after the next line? 😅

mellinoe:
I've actually found an existing problem, which is a bit tricky to fix. It affects this scenario. Apparently, if we want to modify the depth and stencil sizes, we need to do so before a window is created. Unfortunately, it looks like SDL2 is just giving us default depth/stencil values no matter what we specify here, even though the return values indicate success. The SDL documentation actually does say that you need to call these functions before window creation: https://wiki.libsdl.org/SDL_GL_SetAttribute.

It is a bit tricky because VeldridStartup lets you create a Window separately from the GraphicsDevice. Other backends don't care about this at all. We should be able to fix this in `CreateWindowAndGraphicsDevice`, but not the standalone `CreateGraphicsDevice` function.

mellinoe:
That's an existing problem and not made any worse by your change. I'll merge this and look into a real fix for the problem.

mellinoe:
I merged a fix for the above. Note that 32-bit depth formats don't seem to actually bit supported by SDL2 as far as I can tell (there's not much indication as to what you're actually allowed to pass in). I've improved the error detection when context creation fails as well, because sometimes a non-zero context handle is returned even if context creation failed (I have no idea why).

It seems that all of the other depth formats should be supported, though, including the combined depth-stencil formats.

Additionally, specifying no depth buffer (SwapchainDepthFormat == null) now correctly skips creating the depth buffer.

PR #63, opened 2018-04-09
Implement equality and multiplication operators for RgbaFloat & RgbaByte

SomeAnon42:
In addition:
* static Multiply methods
* ToRgbaByte()/ToRgbaFloat()
* RgbaFloat now has a public constructor that accepts a Vector4

### DIFF
--- a/src/Veldrid/RgbaByte.cs
+++ b/src/Veldrid/RgbaByte.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Runtime.CompilerServices;
 
 namespace Veldrid
 {
@@ -101,6 +102,7 @@ namespace Veldrid
         /// </summary>
         /// <param name="other">The instance to compare to.</param>
         /// <returns>True if all elements are equal; false otherswise.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public bool Equals(RgbaByte other)
         {
             return R.Equals(other.R) && G.Equals(other.G) && B.Equals(other.B) && A.Equals(other.A);
@@ -110,6 +112,7 @@ namespace Veldrid
         /// Returns the hash code for this instance.
         /// </summary>
         /// <returns>A 32-bit signed integer that is the hash code for this instance.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public override int GetHashCode()
         {
             return HashHelper.Combine(R.GetHashCode(), G.GetHashCode(), B.GetHashCode(), A.GetHashCode());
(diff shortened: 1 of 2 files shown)


mellinoe:
Looks reasonable. Should there be multiplication operators and methods for both parameters orders?

SomeAnon42:
Indeed, there should be.

SomeAnon42:
Done.

src/Veldrid/RgbaByte.cs
```diff
@@ -123,5 +133,70 @@ public override string ToString()
         {
             return string.Format("R:{0}, G:{1}, B:{2}, A:{3}", R, G, B, A);
         }
+
+        /// <summary>
+        /// Multiplies the channels of an RgbaByte by the given value.
+        /// </summary>
+        /// <param name="value">The source RgbaByte.</param>
+        /// <param name="scale">The value to multiply by.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
mellinoe: We should either remove this or apply it to all the new methods, otherwise it's inconsistent.

SomeAnon42: I think applying it to the other methods won't hurt.

|                    Method |     Mean |     Error |    StdDev | Scaled |
|-------------------------- |---------:|----------:|----------:|-------:|
|         RgbaByte_Multiply | 28.39 ns | 0.0952 ns | 0.0844 ns |   1.00 |
| RgbaByte_Multiply_Inlined | 20.21 ns | 0.1822 ns | 0.1521 ns |   0.71 |

``` ini

BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17133.73 (1803/SpringCreatorsUpdate/Redstone4)
Intel Core i5-3210M CPU 2.50GHz (Ivy Bridge), 1 CPU, 4 logical and 2 physical cores
Frequency=2435886 Hz, Resolution=410.5282 ns, Timer=TSC
.NET Core SDK=2.1.300-preview1-008174
  [Host]     : .NET Core 2.0.6 (CoreCLR 4.6.26212.01, CoreFX 4.6.26212.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.0.6 (CoreCLR 4.6.26212.01, CoreFX 4.6.26212.01), 64bit RyuJIT

```

mellinoe: Sounds good to me.

src/Veldrid/RgbaByte.cs
```diff
@@ -123,5 +136,75 @@ public override string ToString()
...
+        public static RgbaByte Multiply(RgbaByte value, float scale)
+        {
+            return new RgbaByte(
```
mellinoe: This has bad behavior in cases where channels overflow. RgbaFloat has slightly better semantics since it uses floats, but at least that domain is better understood for representing colors. If you overflow an RgbaByte, your color is basically meaningless. If you clamp to (255, 255, 255, 255), then the scaling becomes skewed. I'm not really sure how to handle this other than to remove the stuff from RgbaByte...

SomeAnon42: MonoGame uses clamp apparently...

src/Veldrid/RgbaByte.cs
```diff
@@ -96,11 +97,22 @@ public RgbaByte(byte r, byte g, byte b, byte a)
...
         }
 
+        /// <summary>
+        /// Converts this RgbaByte into an RgbaFloat.
+        /// </summary>
+        /// <returns></returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public RgbaFloat ToRgbaFloat()
+        {
+            return new RgbaFloat(R / 255.0f, G / 255.0f, B / 255.0f, A / 255.0f);
```
mellinoe: This also has bad overflow behavior.

SomeAnon42:
@mellinoe I'm thinking maybe the equality operators and a public constructor accepting a vector is all we need. Anyone can implement the rest if needed (as extension methods).

SomeAnon42:
@mellinoe Removed the questionable changes and squashed the commits.

mellinoe:
We should also override Object.Equals(object o):

> warning CS0660: 'RgbaByte' defines operator == or operator != but does not override Object.Equals(object o)
> warning CS0660: 'RgbaFloat' defines operator == or operator != but does not override Object.Equals(object o)

mellinoe:
I've manually merged this, and added the Equals(object) overrides as well. Thanks!

PR #67, opened 2018-04-27
Improve OpenGL uniform buffer size error message

tgjones:


mellinoe:
This is a good improvement. I think we can also give the name of the resource, since it's stored in the OpenGLResourceLayout referenced by the OpenGLResourceSet:

```C#
string name = glResourceSet.Layout.Elements[element].Name;
```

src/Veldrid/OpenGL/OpenGLCommandExecutor.cs
```diff
@@ -654,7 +654,7 @@ public void SetComputeResourceSet(uint slot, ResourceSet rs)
                             if (glUB.SizeInBytes < uniformBindingInfo.BlockSize)
                             {
                                 throw new VeldridException(
-                                    $"Not enough data in uniform buffer. Shader expects at least {uniformBindingInfo.BlockSize}, but buffer only contains {glUB.SizeInBytes}");
+                                    $"Not enough data in uniform buffer in slot {slot}, element {element}. Shader expects at least {uniformBindingInfo.BlockSize}, but buffer only contains {glUB.SizeInBytes}");
```
mellinoe: Let's use the term "set" instead of "slot". The terms are a little bit muddy, but I think that makes it clear that it's talking about the index of the ResourceSet. `element` seems clear enough to me with that context.

Issue #72, opened 2018-05-09
Merge branch "sdl2-mouse-additions"


erictuvesson:
Hey, could you merge the "sdl2-mouse-additions" branch? 
I would really like to have the mouse functionality.

mellinoe:
Sure, I think it's pretty useful as well. It would be nice if there was also a way to get the "Mouse Delta" rather than just the absolute position, and perhaps that delta could ignore changes caused by SetMousePosition. With just SetMousePosition, you'll have to keep track of that separately.

mellinoe:
I merged https://github.com/mellinoe/veldrid/commit/6394869af0477d259922541325357f8b5303e82b, so SetMousePosition is now available.

PR #73, opened 2018-05-09
Add MouseDelta to InputSnapshot

erictuvesson:
You talked about it in #72

### DIFF
--- a/src/Veldrid.SDL2/Sdl2Window.cs
+++ b/src/Veldrid.SDL2/Sdl2Window.cs
@@ -241,6 +241,8 @@ namespace Veldrid.Sdl2
             if (_exists)
             {
                 SDL_WarpMouseInWindow(_window, x, y);
+                _currentMouseX = x;
+                _currentMouseY = y;
             }
         }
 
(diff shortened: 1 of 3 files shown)


mellinoe:
This looks good, code-wise. The reason I didn't add it is because I'm not sure how to handle the `InputSnapshot` change. Technically, it is a breaking change to add a new property to an interface. I need to think about how we could cleanly add this...

src/Veldrid.SDL2/Sdl2Window.cs
```diff
@@ -545,6 +547,7 @@ private MouseButton MapMouseButton(SDL_MouseButton button)
         private void HandleMouseMotionEvent(SDL_MouseMotionEvent mouseMotionEvent)
         {
             Vector2 mousePos = new Vector2(mouseMotionEvent.x, mouseMotionEvent.y);
+            _privateSnapshot.MouseDelta = new Vector2(_currentMouseX - (int)mousePos.X, _currentMouseY - (int)mousePos.Y);
```
erictuvesson: Don't need to cast the mousePos here because SDL _should_ return integers.

mellinoe:
@erictuvesson Sorry about the long delay on this. Instead of adding a new property to InputSnapshot (which would be a breaking change), I've instead added the property directly into the Sdl2Window class. When `PumpEvents` is called, the MouseDelta property will be updated.

PR #78, opened 2018-05-15
Add a boolean that tells about the UV coordinate origin

feliwir:


tgjones:
It’s a finickity point, but could this property be named “UvStartsAtTop”? I think it would read better.

mellinoe:
Looks good. As for naming, I think `UvStartsAtTop` is a little better. I was going to suggest `UvOriginTopLeft`, but I'm not sure I have a clear preference on that. Thoughts?

Also, let's augment the XML comment to:

> Gets a value identifying whether texture coordinates begin in the top left corner of a Texture. If true, (0, 0) refers to the top-left texel of a Texture. If false, (0, 0) refers to the bottom-left texel of a Texture. This property is useful when determining how the output of a Framebuffer should be sampled.

feliwir:
Then I'd probably prefer UvOriginTopLeft since that's more explicit

mellinoe:
I went with `IsUvOriginTopLeft`, in the end.

PR #86, opened 2018-06-03
Rewrite StagingMemoryPool to use unmanaged memory blocks

SomeAnon42:
This change noticeably improves performance (mainly) in scenarios that involve lots of texture updates, such as video playback.

A few things to note:
- The pool returns *the smallest* block that is big enough to store the requested amount of data.
- ``GraphicsDevice`` now owns (the only instance of) the pool.
- ``StagingMemoryPool`` now has a ``Dispose`` method which frees all the native memory. The execution thread is responsible for calling it.
- Each ``CommandList`` has its own list of staging blocks, which is flushed when the ``CommandList`` is disposed. This presents a problem if a ``CommandList`` is disposed after the ``GraphicsDevice``, which owns the memory pool. This is currently addressed by adding [this check](https://github.com/SomeAnon42/veldrid/blob/c6f73d4e11dfbbf825adff2e8b91c4e6dddb160e/src/Veldrid/StagingMemoryPool.cs#L82). Essentially, when the pool is disposed (and the list of staging blocks is empty), calling ``Free`` results in a no-op.
- ``UpdateTexture`` no longer allocates an ``Action`` and a closure object. A staging block is used to pass the arguments to the execution thread. We could go even further and change ``OpenGLCommandExecutor.UpdateTexture`` to accept a reference to ``UpdateTextureArgs`` instead of separate parameters to avoid extra copying.
- I'm not really sure where to put ``UpdateTextureArgs``and whether it should be a nested struct.

### DIFF
--- a/src/Veldrid/OpenGL/NoAllocEntryList/NoAllocUpdateBufferEntry.cs
+++ b/src/Veldrid/OpenGL/NoAllocEntryList/NoAllocUpdateBufferEntry.cs
@@ -4,10 +4,10 @@
     {
         public readonly Tracked<DeviceBuffer> Buffer;
         public readonly uint BufferOffsetInBytes;
-        public readonly Tracked<byte[]> StagingBlock;
+        public readonly StagingBlock StagingBlock;
         public readonly uint StagingBlockSize;
 
-        public NoAllocUpdateBufferEntry(Tracked<DeviceBuffer> buffer, uint bufferOffsetInBytes, Tracked<byte[]> stagingBlock, uint stagingBlockSize)
+        public NoAllocUpdateBufferEntry(Tracked<DeviceBuffer> buffer, uint bufferOffsetInBytes, StagingBlock stagingBlock, uint stagingBlockSize)
         {
             Buffer = buffer;
             BufferOffsetInBytes = bufferOffsetInBytes;
--- a/src/Veldrid/OpenGL/OpenGLGraphicsDevice.cs
+++ b/src/Veldrid/OpenGL/OpenGLGraphicsDevice.cs
@@ -91,7 +91,6 @@ namespace Veldrid.OpenGL
 
         public StagingMemoryPool StagingMemoryPool => _stagingMemoryPool;
 
-
         public OpenGLGraphicsDevice(
             GraphicsDeviceOptions options,
             OpenGLPlatformInfo platformInfo,
@@ -165,7 +164,7 @@ namespace Veldrid.OpenGL
                 texture1D: _backendType == GraphicsBackend.OpenGL,
                 independentBlend: _extensions.IndependentBlend,
                 structuredBuffer: _extensions.StorageBuffers,
-                subsetTextureView: _extensions.ARB_TextureView );
+                subsetTextureView: _extensions.ARB_TextureView);
 
             _resourceFactory = new OpenGLResourceFactory(this);
 
@@ -773,7 +772,6 @@ namespace Veldrid.OpenGL
             StagingBlock textureData = _stagingMemoryPool.Stage(source, sizeInBytes);
             StagingBlock argBlock = _stagingMemoryPool.GetStagingBlock(UpdateTextureArgsSize);
             ref UpdateTextureArgs args = ref Unsafe.AsRef<UpdateTextureArgs>(argBlock.Data);
-            args.Texture = texture;
             args.Data = (IntPtr)textureData.Data;
             args.X = x;
             args.Y = y;
@@ -784,14 +782,13 @@ namespace Veldrid.OpenGL
             args.MipLevel = mipLevel;
             args.ArrayLayer = arrayLayer;
 
-            _executionThread.UpdateTexture(argBlock.Id, textureData.Id);
+            _executionThread.UpdateTexture(texture, argBlock.Id, textureData.Id);
         }
 
-        public static readonly uint UpdateTextureArgsSize = (uint)Unsafe.SizeOf<UpdateTextureArgs>();
+        private static readonly uint UpdateTextureArgsSize = (uint)Unsafe.SizeOf<UpdateTextureArgs>();
 
-        public struct UpdateTextureArgs
+        private struct UpdateTextureArgs
         {
-            public Texture Texture;
             public IntPtr Data;
             public uint X;
             public uint Y;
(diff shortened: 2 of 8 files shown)


src/Veldrid/StagingMemoryPool.cs
```diff
@@ -1,114 +1,133 @@
...
+            return block;
         }
 
-        public FixedStagingBlock GetFixedStagingBlock(uint sizeInBytes)
+        public StagingBlock GetStagingBlock(uint sizeInBytes)
         {
-            byte[] array = _arrayPool.Rent((int)sizeInBytes);
-            return new FixedStagingBlock(array, sizeInBytes, this);
+            Rent(sizeInBytes, out StagingBlock block);
+            Unsafe.InitBlock(block.Data, 0, sizeInBytes);
```
mellinoe: Is clearing the data actually necessary?

src/Veldrid/StagingMemoryPool.cs
```diff
@@ -1,114 +1,133 @@
...
-    internal unsafe struct StagingBlock
-    {
-        public readonly byte[] Array;
-        public readonly uint SizeInBytes;
-        public readonly StagingMemoryPool Pool;
+        public void Free(StagingBlock block)
+        {
+            lock (_lock)
+            {
+                if (block.Id < _storage.Count)
```
mellinoe: It feels like this should be an assertion or something. What happens when a block with a nonsense ID is returned?

src/Veldrid/StagingMemoryPool.cs
```diff
@@ -1,114 +1,133 @@
...
 
-        public void Free(byte[] array)
+        private void Allocate(uint sizeInBytes, out StagingBlock stagingBlock)
         {
-            bool clearArray = false;
-#if DEBUG
-            clearArray = true;
-#endif
-            _arrayPool.Return(array, clearArray);
+            IntPtr ptr = Marshal.AllocHGlobal((int)sizeInBytes);
```
mellinoe: I think we should have a minimum allocation size. Perhaps 64 or 128 bytes.

src/Veldrid/StagingMemoryPool.cs
```diff
@@ -1,114 +1,133 @@
...
+                for (int i = 0; i < available.Count; i++)
+                {
+                    int index = (int)indices[i];
+                    StagingBlock current = _storage[index];
+                    if (current.Capacity >= size)
+                    {
+                        available.RemoveAt(i);
+                        current.SizeInBytes = size;
+                        block = current;
+                        _storage[index] = current;
```
mellinoe: Is it necessary to store the block back? I know "SizeInBytes" was changed, but that doesn't seem important to remember in that list.

src/Veldrid/OpenGL/OpenGLGraphicsDevice.cs
```diff
@@ -88,6 +89,9 @@ public override bool SyncToVerticalBlank
 
         public override GraphicsDeviceFeatures Features => _features;
 
+        public StagingMemoryPool StagingMemoryPool => _stagingMemoryPool;
+
```
mellinoe: nit: One extra newline

src/Veldrid/OpenGL/OpenGLGraphicsDevice.cs
```diff
@@ -766,15 +770,37 @@ protected override void UpdateBufferCore(DeviceBuffer buffer, uint bufferOffsetI
...
+            args.Width = width;
+            args.Height = height;
+            args.Depth = depth;
+            args.MipLevel = mipLevel;
+            args.ArrayLayer = arrayLayer;
+
+            _executionThread.UpdateTexture(argBlock.Id, textureData.Id);
+        }
+
+        public static readonly uint UpdateTextureArgsSize = (uint)Unsafe.SizeOf<UpdateTextureArgs>();
```
mellinoe: This can probably be private

src/Veldrid/OpenGL/OpenGLGraphicsDevice.cs
```diff
@@ -766,15 +770,37 @@ protected override void UpdateBufferCore(DeviceBuffer buffer, uint bufferOffsetI
...
+
+        public struct UpdateTextureArgs
```
mellinoe: Same, it can probably be private

src/Veldrid/OpenGL/OpenGLGraphicsDevice.cs
```diff
@@ -955,100 +981,113 @@ private void ExecuteWorkItem(ExecutionThreadWorkItem workItem)
...
+                                    stagingBlock.SizeInBytes);
+
+                                stagingBlock.Free();
                             }
-                            pool.Free(stagingArray);
-                        }
-                        break;
+                            break;
+                        case WorkItemType.UpdateTexture:
+                            unsafe
```
mellinoe: Isn't this whole class already unsafe?

src/Veldrid/StagingMemoryPool.cs
SomeAnon42: No, it actually is important. You don't want ``RetrieveById`` to return a block with an incorrect size.

src/Veldrid/StagingMemoryPool.cs
SomeAnon42: Seems reasonable.

src/Veldrid/StagingMemoryPool.cs
SomeAnon42: >What happens when a block with a nonsense ID is returned
Nothing, and that's the whole point. I added this check so that calling ``Free`` wouldn't have any effect after the pool is disposed (and therefore _storage.Count is 0).

src/Veldrid/StagingMemoryPool.cs
```diff
@@ -1,114 +1,133 @@
...
+            return block;
         }
 
         public StagingBlock Stage(byte[] bytes)
         {
-            byte[] array = _arrayPool.Rent(bytes.Length);
-            Array.Copy(bytes, array, bytes.Length);
-            return new StagingBlock(array, (uint)bytes.Length, this);
+            Rent((uint)bytes.Length, out StagingBlock block);
+            Marshal.Copy(bytes, 0, (IntPtr)block.Data, bytes.Length);
```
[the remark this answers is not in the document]
SomeAnon42: Got it.

src/Veldrid/OpenGL/OpenGLGraphicsDevice.cs
SomeAnon42: Indeed.

src/Veldrid/OpenGL/OpenGLGraphicsDevice.cs
SomeAnon42: I think I left it public because I was experimenting with this idea
>We could go even further and change OpenGLCommandExecutor.UpdateTexture to accept a reference to UpdateTextureArgs instead of separate parameters to avoid extra copying

src/Veldrid/StagingMemoryPool.cs
SomeAnon42: If you're worried about copying, we could store the blocks in an array and then use ``ref`` to update or return them.

src/Veldrid/StagingMemoryPool.cs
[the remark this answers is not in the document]
SomeAnon42: Actually... ``Buffer.MemoryCopy`` accepts a pointer as the source, and here we have a managed array.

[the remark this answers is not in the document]
mellinoe: Gotcha -- this is fine then.

src/Veldrid/StagingMemoryPool.cs
mellinoe: Understood. It might be worth adding an assertion that the pool has been disposed if we hit this case, just to make it clear what the expectation is. Or, at least a comment of some sort. At first glance it's not clear what the check is for.

src/Veldrid/StagingMemoryPool.cs
mellinoe: > No, it actually is important. You don't want RetrieveById to return a block with an incorrect size.

That makes sense. Ignore this one.

src/Veldrid/OpenGL/OpenGLGraphicsDevice.cs
```diff
@@ -955,100 +981,113 @@ private void ExecuteWorkItem(ExecutionThreadWorkItem workItem)
...
-                            OpenGLCommandEntryList list = (OpenGLCommandEntryList)workItem.Object0;
-                            try
                             {
-                                list.ExecuteAll(_gd._commandExecutor);
-                                list.Parent.OnCompleted(list);
-                            }
-                            finally
-                            {
-                                if (!_gd.CheckCommandListDisposal(list.Parent))
+                                OpenGLCommandEntryList list = (OpenGLCommandEntryList)workItem.Object0;
```
mellinoe: The indentation of these "case" blocks got changed in the PR -- can you revert that? There's an option in VS for not indenting blocks inside cases:

image

SomeAnon42: Weird, I couldn't see these changes in VS (and I still can't)

mellinoe: The blocks under the `case` statements are indented, when they shouldn't be. If you uncheck that option, then auto-formatting the document should fix it (does for me, at least)

src/Veldrid/OpenGL/NoAllocEntryList/NoAllocUpdateBufferEntry.cs
```diff
@@ -4,10 +4,10 @@ internal struct NoAllocUpdateBufferEntry
     {
         public readonly Tracked<DeviceBuffer> Buffer;
         public readonly uint BufferOffsetInBytes;
-        public readonly Tracked<byte[]> StagingBlock;
+        public readonly StagingBlock StagingBlock;
```
mellinoe: There's a subtle problem with this -- `StagingBlock` contains a managed reference (to the pool), but `NoAllocUpdateBufferEntry` instances are stored in a non-GC-tracked unsafe buffer. This could potentially introduce GC holes, although it's very unlikely since the lifetime of these entries should be strictly within the lifetime of the GraphicsDevice anyways.

To fix this, we should remove the StagingMemoryPool reference from StagingBlock, and pass the pool in to the Free method.

mellinoe:
I've been debugging through this, trying to make sure there's no leaks. I came across a pre-existing problem with OpenGLShader that causes memory to leak when a shader is disposed before it is initialized. The staging block is never freed in that case. I'll paste my fix below. Feel free to include it in your changeset, or I will add it myself after the eventual merge.

OpenGLShader:
```C#
        public void DestroyGLResources()
        {
            if (!_disposed)
            {
                _disposed = true;
                if (Created)
                {
                    glDeleteShader(_shader);
                    CheckLastError();
                }
                else
                {
                    _stagingBlock.Free();
                }
            }
        }
```

mellinoe:
Thanks for the contribution. I think everything looks good to merge now.

PR #87, opened 2018-06-06
Update cached position on move

barrybingo:
_cachedPosition only gets updated via SDL_WindowEventID.Moved
if try to change both X and Y then second call to SetWindowPosition will use old cached value

Example
```c#
Sdl2Window _window;
_window.X = newX; // SetWindowPosition(newX, _cachedPosition.Value.Y)
_window.Y = newY; // SetWindowPosition(_cachedPosition.Value.X, newY) // BAD

```
Maybe and/or make Sdl2Window.SetWindowPosition public ?

### DIFF
--- a/src/Veldrid.SDL2/Sdl2Window.cs
+++ b/src/Veldrid.SDL2/Sdl2Window.cs
@@ -930,6 +930,7 @@ namespace Veldrid.Sdl2
         private void SetWindowPosition(int x, int y)
         {
             SDL_SetWindowPosition(_window, x, y);
+            _cachedPosition.Value = new Point(x, y);
         }
 
         private Point GetWindowSize()


mellinoe:
Thanks, this is a good fix. I'll merge now.

> Sdl2Window.SetWindowPosition

Maybe it would be better to just have a get/set `Position` property? Either way, I don't see a problem with that. It will be a little bit more convenient than what we have currently.

barrybingo:
X and Y could be get only with a get/set Poistion property.
Do same with Width and Height.

barrybingo:
X and Y could be get only with a get/set Position property.

Do same with Width and Height.

mellinoe:
I won't make any breaking changes (like removing properties), but adding a new ones for Position and Size seem okay to me.

PR #89, opened 2018-06-10
OpenGL Context is now requested with ForwardCompatible Flag

AlzPatz:
Suggested change, please modify any way you see fit. Thank you

MESA drivers on Linux distros do not appear to return contexts higher than OpenGL 3.0 unless a Forward Compatible (no depreciated features) Core context is requested. The latest MESA drivers appear to sometimes not require this fix, but this improved behaviour appears unrealiabe under my limited testing.


mellinoe:

AlzPatz:
No probs - hopefully I can contribute more in the future. I’ll watch out for any items that need some attention. I need to get more up to speed with the build process and the interop code but hopefully there is somewhere I can be useful

PR #96, opened 2018-07-06
[Vulkan] Avoid hidden enumerator allocations

SomeAnon42:
Saw this when profiling allocations
image

image


mellinoe:
Ouch, that's an obvious one. Thanks for the fix.

PR #105, opened 2018-07-25
Avoid allocations when calling GetHashCode or Equals on certain structs

SomeAnon42:
* Replace GetHashCode/Equals calls on enums with casts to int.
* Manually compare nullable values as Nullable does not implement ``IEquatable<T?>``. See ``Util.NullableEquals`` and ``GraphicsPipelineDescription.Equals``.
The BCL has ``Nullable.Equals<T>``, but it performs considerably worse than a custom implementation.

### DIFF
--- a/src/Veldrid/GraphicsPipelineDescription.cs
+++ b/src/Veldrid/GraphicsPipelineDescription.cs
@@ -160,7 +160,9 @@ namespace Veldrid
                 && PrimitiveTopology == other.PrimitiveTopology
                 && ShaderSet.Equals(other.ShaderSet)
                 && Util.ArrayEquals(ResourceLayouts, other.ResourceLayouts)
-                && ResourceBindingModel.Equals(other.ResourceBindingModel)
+                && (ResourceBindingModel.HasValue && other.ResourceBindingModel.HasValue
+                    ? ResourceBindingModel.Value == other.ResourceBindingModel.Value
+                    : ResourceBindingModel.HasValue == other.ResourceBindingModel.HasValue)
                 && Outputs.Equals(other.Outputs);
         }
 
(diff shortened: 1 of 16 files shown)


src/Veldrid/GraphicsPipelineDescription.cs
```diff
@@ -160,7 +160,9 @@ public bool Equals(GraphicsPipelineDescription other)
                 && PrimitiveTopology == other.PrimitiveTopology
                 && ShaderSet.Equals(other.ShaderSet)
                 && Util.ArrayEquals(ResourceLayouts, other.ResourceLayouts)
-                && ResourceBindingModel.Equals(other.ResourceBindingModel)
+                && (ResourceBindingModel.HasValue && other.ResourceBindingModel.HasValue
```
mellinoe: Should this use Util.NullableEquals?

mellinoe:
These look like good fixes to me. There is probably some kind of Roslyn analyzer that I should be using to catch these sorts of unnecessary allocations.

src/Veldrid/GraphicsPipelineDescription.cs
SomeAnon42: It's an enum, and Util.NullableEquals has an IEquatable<T> constraint. We could remove the constraint and use EqualityComparer<T>.Default to compare the values, but that would make it as slow as Nullable.Equals<T> (since that's exactly how it's implemented).

|         Method |      Mean |     Error |    StdDev |
|--------------- |----------:|----------:|----------:|
| CustomEquality |  2.430 ns | 0.0131 ns | 0.0122 ns |
| NullableEquals | 14.788 ns | 0.1120 ns | 0.1048 ns |

It's convenient to use pipeline state as a dictionary key (even though there's probably options that involve less hashing and comparsions), so I think Equals should be as fast as possible.

SomeAnon42:
There's [RoslynClrHeapAllocationAnalyzer](https://github.com/Microsoft/RoslynClrHeapAllocationAnalyzer/releases). It's been updated to support VS 2017.

src/Veldrid/GraphicsPipelineDescription.cs
mellinoe: Okay, that makes sense. I'd like to update to C# 7.3 at some point, but it isn't really worth it for just this one change. This is good as-is.

SomeAnon42:
There is a similar open [PR on the vk repo](https://github.com/mellinoe/vk/pull/20).

mellinoe:
@SomeAnon42 Thanks for the improvement here. I'll move the Vulkan PR along soon and then update the reference here.

PR #111, opened 2018-08-10
[D3D11] Respect InstanceStepRate when creating InputLayoutCacheKey

SomeAnon42:
Currently, input layouts with a non-zero InstanceStepRate aren't properly cached. Requesting the same input layout twice results in an exception. https://github.com/mellinoe/veldrid/blob/d7b6425fe1b502bfefe7eebd4002801a13881eb1/src/Veldrid/D3D11/D3D11ResourceCache.cs#L167

There is a mismatch between the "temporary" key and the "permanent" key in the dictionary, since only the temporary key has the correct InstanceStepRate value. This results in an attempt to add the same InputLayout to the cache under the same (incorrectly generated) key,

Repro:
```csharp
(Shader vs, Shader fs) = LoadShaderSet(factory, "ShaderSetName");
var shaderSetDesc = new ShaderSetDescription(
    new[] { vertexLayout, perInstanceDataLayout }, // has a VertexLayout with a non-0 InstanceStepRate
    new[] { vs, fs });

// ...

var pipelineDesc = new GraphicsPipelineDescription(...);
// Create two pipelines with the same vertex layouts
Pipeline pipeline0 = factory.CreateGraphicsPipeline(ref pipelineDesc);
// Will crash in D3D11ResourceCache.GetInputLayout
Pipeline pipeline1 = factory.CreateGraphicsPipeline(ref pipelineDesc);

### DIFF
--- a/src/Veldrid/D3D11/D3D11ResourceCache.cs
+++ b/src/Veldrid/D3D11/D3D11ResourceCache.cs
@@ -281,6 +281,7 @@ namespace Veldrid.D3D11
                 for (int i = 0; i < original.Length; i++)
                 {
                     vertexLayouts[i].Stride = original[i].Stride;
+                    vertexLayouts[i].InstanceStepRate = original[i].InstanceStepRate;
                     vertexLayouts[i].Elements = (VertexElementDescription[])original[i].Elements.Clone();
                 }
 


mellinoe:
Thanks for the fix. I also merged in a test case for this.

PR #115, opened 2018-08-27
* Skybox.cs: Used new cubemap texture class to remove unsafe code

geoeo:
* ImageSharpCubeMapTexture.cs: Added new cubemap class

Basically a glorified copy paste of the code in skybox.cs
At the moment cubemap textures can not be mipmapped.
I was not sure how to deal with this that fast. 
Also, opengl only allows cubemap mipmaps with version >= 4.0


src/Veldrid.ImageSharp/ImageSharpCubeMapTexture.cs
```diff
@@ -0,0 +1,105 @@
...
+                gd.UpdateTexture(cubeMapTexture, (IntPtr)bottomPin, faceSize, 0, 0, 0, Width, Height, 1, 0, 3);
+                gd.UpdateTexture(cubeMapTexture, (IntPtr)backPin, faceSize, 0, 0, 0, Width, Height, 1, 0, 4);
+                gd.UpdateTexture(cubeMapTexture, (IntPtr)frontPin, faceSize, 0, 0, 0, Width, Height, 1, 0, 5);
+
+            }
+
+            return cubeMapTexture;
+
+        }
+
```
mellinoe: nit: Extra newline

src/Veldrid.ImageSharp/ImageSharpCubeMapTexture.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// <summary>
+        /// The size of each pixel, in bytes.
+        /// </summary>
+        public uint PixelSizeInBytes => sizeof(byte) * 4;
+
+        /// <summary>
+        /// Provides standardized access to the cubemap texture array
+        /// </summary>
+        private const int FRONT = 0;
```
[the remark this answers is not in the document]
mellinoe: Additionally, since these are the order they appear in the public array, we should make these values correspond to the texture array layers that they end up in. The order is:

+X
-X
+Y
-Y
+Z
-Z

Perhaps those are better names for these constants anyways, since front/back implies a handedness convention.

src/Veldrid.ImageSharp/ImageSharpCubeMapTexture.cs
```diff
@@ -0,0 +1,105 @@
...
+        private const int BACK = 1;
+        private const int LEFT = 2;
+        private const int RIGHT = 3;
+        private const int TOP = 4;
+        private const int BOTTOM = 5;
+
+        public ImageSharpCubeMapTexture(string frontPath,
```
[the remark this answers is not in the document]
mellinoe: I think we should rename these so that they don't mention direction (since that is based on coordinate system conventions). Instead, they could be named `positiveXPath`, `negativeXPath`, `positiveYPath`, etc.

src/Veldrid.ImageSharp/ImageSharpCubeMapTexture.cs
```diff
@@ -0,0 +1,105 @@
+﻿using SixLabors.ImageSharp;
+using SixLabors.ImageSharp.Advanced;
+using SixLabors.ImageSharp.PixelFormats;
+using System;
+using System.Runtime.InteropServices;
+
+namespace Veldrid.ImageSharp
+{
+    public class ImageSharpCubeMapTexture
```
mellinoe: Other parts of Veldrid's documentation spell this as "Cubemap" ("map" not being capitalized). For consistency, we should do that here.

src/Veldrid.ImageSharp/ImageSharpCubeMapTexture.cs
```diff
@@ -0,0 +1,105 @@
...
+using SixLabors.ImageSharp.Advanced;
+using SixLabors.ImageSharp.PixelFormats;
+using System;
+using System.Runtime.InteropServices;
+
+namespace Veldrid.ImageSharp
+{
+    public class ImageSharpCubeMapTexture
+    {
+
```
mellinoe: nit: Extra newline

src/Veldrid.ImageSharp/ImageSharpCubemapTexture.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// </summary>
+        private const int POSITIVE_X = 0;
+        private const int NEGATIVE_X = 1;
+        private const int POSITIVE_Y = 2;
+        private const int NEGATIVE_Y = 3;
+        private const int POSITIVE_Z = 4;
+        private const int NEGATIVE_Z = 5;
+
+        public ImageSharpCubemapTexture(
+            string negativeZPath,
```
mellinoe: We should re-order these so that they are in the standard array-layer order (the order that the constants are in). The stuff in NeoDemo will also need to be re-ordered.

src/Veldrid.ImageSharp/ImageSharpCubemapTexture.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        /// <summary>
+        /// The size of each pixel, in bytes.
+        /// </summary>
+        public uint PixelSizeInBytes => sizeof(byte) * 4;
+
+        /// <summary>
+        /// Provides standardized access to the cubemap texture array
+        /// </summary>
+        private const int POSITIVE_X = 0;
```
mellinoe: style/naming: Please switch these to Pascal case, with a mention of "ArrayLayer", e.g.:

```C#
private const int PositiveXArrayLayer = 0;
private const int NegativeXArrayLayer = 1;
// etc.
```

geoeo:
About mipmapping. Since mipmapping cubemaps is only supported for OpenGL >= 4.0, it will throw an exception on mac (I think). Maybe I misunderstood something: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGenerateMipmap.xhtml

How can I query the OS level via Veldrid, so I can return false if mipmapping is requested on a mac with opengl backend.

src/Veldrid.ImageSharp/ImageSharpTexture.cs
```diff
@@ -137,7 +137,7 @@ private unsafe Texture CreateTextureViaUpdate(GraphicsDevice gd, ResourceFactory
             return tex;
         }
 
-        private static Image<T>[] GenerateMipmaps<T>(Image<T> baseImage) where T : struct, IPixel<T>
+        public static Image<T>[] GenerateMipmaps<T>(Image<T> baseImage) where T : struct, IPixel<T>
```
mellinoe: This only needs to be internal, rather than public.

src/Veldrid.ImageSharp/ImageSharpCubemapTexture.cs
```diff
@@ -0,0 +1,151 @@
...
+                CubemapTextures[1] = new Image<Rgba32>[1] { negativeX };
+                CubemapTextures[2] = new Image<Rgba32>[1] { positiveY };
+                CubemapTextures[3] = new Image<Rgba32>[1] { negativeY };
+                CubemapTextures[4] = new Image<Rgba32>[1] { positiveZ };
+                CubemapTextures[5] = new Image<Rgba32>[1] { negativeZ };
+            }
+
+        }
+
+        public unsafe Texture CreateCubemapTextures(GraphicsDevice gd, ResourceFactory factory)
```
mellinoe: I think this should be called "CreateDeviceTexture", like the other class.

src/Veldrid.ImageSharp/ImageSharpCubemapTexture.cs
```diff
@@ -0,0 +1,151 @@
...
+        { }
+
+        public ImageSharpCubemapTexture(
+            Image<Rgba32> positiveX,
+            Image<Rgba32> negativeX,
+            Image<Rgba32> positiveY,
+            Image<Rgba32> negativeY,
+            Image<Rgba32> positiveZ,
+            Image<Rgba32> negativeZ,
+            bool mipmap = false)
```
mellinoe: I think the default for this should be "true" as it is in ImageSharpTexture -- mainly for consistency. NeoDemo should pass false, since a skybox doesn't need mipmaps.

geoeo:
Btw. What do you think of moving GenerateMipmaps from ImageSharpTexture into MipmapHelper?

mellinoe:
@geoeo That makes sense to me, now that there's two classes using it.

mellinoe:
@geoeo This looks good to me now. Thanks for the contribution!

PR #116, opened 2018-09-02
Added an UpdateBuffer function which copies a range of data from an array

geoeo:
Copying data via UpdateBuffer only offers the option of a range copy via IntPtr. 
This methods serves as a way to remove the need to create an IntPtr explicitly.

### DIFF
--- a/src/Veldrid/GraphicsDevice.cs
+++ b/src/Veldrid/GraphicsDevice.cs
@@ -565,6 +565,33 @@ namespace Veldrid
 
         /// <summary>
         /// Updates a <see cref="DeviceBuffer"/> region with new data.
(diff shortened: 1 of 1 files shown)


mellinoe:
You can use this existing overload to accomplish this: https://github.com/mellinoe/veldrid/blob/490cdffd39cb3d362d733001d0440ebde4ee3e96/src/Veldrid/GraphicsDevice.cs#L534-L538

The only difference is you pass the size in bytes, rather than the element count. On top of that, it is more flexible because you can upload data from any arbitrary chunk in the array, not just from the beginning.

geoeo:
Perfect. Thanks!

PR #120, opened 2018-09-16
Added the ability to create centered window

Scellow:


mellinoe:
Instead of adding an entirely new overload, I think it makes more sense to just expose the `SDL_WINDOWPOS_CENTERED` constant, with a more friendly name.

```C#
public static unsafe partial class Sdl2Native
{
    /// <summary>
    /// A special sentinel value indicating that a newly-created window should be centered in the screen.
    /// </summary>
    public const int WindowPositionCentered = 0x2FFF0000;
}
```

Then, you can just use the existing overload, but pass Sdl2Native.WindowPositionCentered for the width and height.

mellinoe:
I went ahead and exposed a new constant, `Sdl2Native.SDL_WINDOWPOS_CENTERED` which can be used to center the window.

PR #129, opened 2018-10-15
Add InitializeSDL to VeldridStartup

AridTag:
I wanted to be able to initialize SDLs context attributes the same way it's done in VeldridStartup currently but it was private.

Figured I would extract and expose a method that allows me to initialize SDL in one call while leveraging the other work that exists in VeldridStartup

### DIFF
--- a/src/Veldrid.StartupUtilities/VeldridStartup.cs
+++ b/src/Veldrid.StartupUtilities/VeldridStartup.cs
@@ -9,6 +9,21 @@ namespace Veldrid.StartupUtilities
 {
     public static class VeldridStartup
     {
+        /// <summary>
+        /// Initialize SDL with SDLInitFlags.Video
+        /// <para>
+        /// If preferredBackend is OpenGL or OpenGLES sets the SDL contexts attributes appropriately
+        /// </para>
+        /// </summary>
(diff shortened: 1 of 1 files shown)


mellinoe:
Hey @AridTag, thanks for the submission! Could you provide a little more information regarding how you would use this, and what scenario it would help you with?

AridTag:
Currently the only way that `SetSDLGLContextAttributes` is exposed is via `CreateWindowAndGraphicsDevice` but I'm creating the window myself and using the `Sdl2Window` constructor that takes a window handle. Then only using `CreateGraphicsDevice`.

I just wanted to be able to initialize SDL and set the GL context attributes without having to go and re-implement what's already being done here.

Hope that helps explain why I did this a little better.

mellinoe:
I see. Would it be reasonable to just expose `SetSDLGLContextAttributes`, in that case? It seems like that is the "core functionality" you are after.

AridTag:
I can live with that. I'll add another commit to this PR later

AridTag:
I put CreateWindowAndGraphicsDevice back to how it was and instead exposed SetSDLGLContextAttributes by itself

PR #137, opened 2018-11-22
Allow non-multiple-of-four-sized texture to be updated with multiple-of-four mipmap data

tgjones:
I found this bug because one SAGE texture is BC1-compressed, but sized 2 x 512. The mipmap data still needs to be multiple-of-four-sized, but this isn't currently allowed by Veldrid.


mellinoe:
Looks good to me. Thanks for fixing this.

PR #141, opened 2018-11-30
Fix D3D11 TextureView creation for arrayLayer > 0

tgjones:
I assume this was a typo - at the moment, it's impossible to create a `TextureView` for anything other than the first array layer in a `Texture2D`  (assuming it has `ArrayLayers > 1`, and assuming you want all mip levels in the texture view).


tgjones:
Oh, also it was impossible to create a `TextureView` for anything other than the first mipmap level, if it was a non-array texture.

mellinoe:
Thanks for catching this. I have the same fix on a local branch, but it has not been merged back upstream yet. I'll double check this in a bit and merge it in.

mellinoe:
@tgjones Looks good to me, thanks for the fix! Just a heads up -- my in-progress branch has a bunch of TextureView changes that will end up moving this and other logic around, and potentially removing this particular section. Until that work gets merged in, though, this is definitely the correct fix for the current logic.

PR #145, opened 2018-12-22
Added: method to load images from streams

equinox2k:


mellinoe:
Looks like a reasonable addition to me. Thanks @equinox2k !

PR #146, opened 2018-12-23
Wayland Vulkan swapchain support

Jjagg:
The first commit is very straightforward, adding the required structs/classes/extensions/functions for supporting Wayland.

If I'm not misunderstanding anything, we only need to enable surface extensions when a swapchain source is provided. The relevant changes for that are in my second commit.

Under XWayland (when SDL acts as an X client, but the host DE uses Wayland), SDL returns `Unknown` for the subsytem in its SysWMinfo, so that doesn't work with the `SwapchainSurface` API. The OpenGL path in Veldrid.StartupUtilities does work on XWayland.


mellinoe:
This is fantastic @Jjagg , thanks for the submission!

Everything functionally looks good, although I will need to figure out how to test this on my Linux box. Is there a way to easily test this with Ubuntu 16.04? I can probably update it if necessary, but it'll have to wait til this weekend when I get back from the holidays.

The one thing I will probably want to change is the extension handling. At the moment, the platform surface extensions are always enabled when running on that platform. This means that if you create a headless device (e.g. without a SwapchainDescription), then you can still later create a Swapchain. I think we should continue on that path and unconditionally enable all possible surface extensions for your platform, regardless of whether you have a matching SwapchainSource ready. The errors should therefore be moved into the logic that actually tries to create a Swapchain from that kind of source instead of at device creation time. So you won't get an error about the Wayland extension being missing unless you try to create a Wayland Swapchain. If you've passed in a Wayland SwapchainSource, then you will still get the exception at device creation time.

Question for my curiosity: is it ever reasonable or possible to create both an X Window and a Wayland window in the same process? It wouldn't change the approach above, but I am interested to know if that is something anyone would ever do.

Jjagg:
> Is there a way to easily test this with Ubuntu 16.04?

Yes, you can use the reference Wayland compositor, `weston`. You can simply install weston with apt and run a nested session by executing `weston` in a terminal.

In the weston session you can open a terminal and run a test project as usual. SDL will by default act as an X client. You can let it use Wayland by setting `SDL_VIDEODRIVER=wayland`. Something like `SDL_VIDEODRIVER=wayland dotnet run` will work. You can tell SDL uses Wayland if the window is not decorated.

> This means that if you create a headless device (e.g. without a SwapchainDescription), then you can still later create a Swapchain. I think we should continue on that path and unconditionally enable all possible surface extensions for your platform.

Yeah, that makes sense. Will change!

> is it ever reasonable or possible to create both an X Window and a Wayland window in the same process?

In general, yes. Wayland compositors can support X clients with XWayland. Imagine an application like ImGui that launches its own window. Maybe an app like that only supports X, but you create your own Window that uses Wayland.
Having both an X and Wayland window using Veldrid is not useful though. If you want to create a window and draw to it, you just want a windowing lib so you can just call `MyLib.CreateWindow` or whatever. That lib (e.g. SDL) will use either X or Wayland. There's no reason to create both X and Wayland windows if you have full control yourself.

Jjagg:
> The errors should therefore be moved into the logic that actually tries to create a Swapchain from that kind of source instead of at device creation time. 

Should we store the enabled surface extensions in the `VkGraphicsDevice` so we can check if the right one is available when an instance is getting created?

mellinoe:
> Should we store the enabled surface extensions in the VkGraphicsDevice so we can check if the right one is available when an instance is getting created?

Yes, I think so. `VkSurfaceUtil.CreateSurface` can take a `VkGraphicsDevice` parameter and it can check whether the relevant surface extension is avaliable / enabled. If not, it can throw an exception.

Thanks for the information about Wayland -- sounds like it will be fairly easy to test once I have my Linux machine in hand.

Jjagg:
@mellinoe `VkSurfaceInfo.CreateSurface` would also need a `VkGraphicsDevice` parameter to pass to `VkSurfaceUtil.CreateSurface`. Is the `VkSurfaceInfo` API still relevant? It seems like it serves the same purpose as `SwapchainSource`, but specifically for Vulkan. There's a [`GraphicsDevice.CreateVulkan`](https://github.com/mellinoe/veldrid/blob/f75915caa500644536601ec4720d9b1f3663043e/src/Veldrid/GraphicsDevice.cs#L994-L1004) overload that takes a `VkSurfaceSource`, but it just converts it to a `SwapchainSource`.
The only other reference is in [`VeldridStartup.GetSurfaceSource`](https://github.com/mellinoe/veldrid/blob/f75915caa500644536601ec4720d9b1f3663043e/src/Veldrid.StartupUtilities/VeldridStartup.cs#L222-L237), but that method isn't used and it's private.

mellinoe:
VkSurfaceInfo Isn’t relevant anymore. That was part of Veldrid 4.0, before Swapchains were a thing. It’s still around for compat, but SwapchainSource is the only thing you need to worry about now.

Jjagg:
How should I resolve this? Adding the `VkGraphicsDevice` parameter breaks compatibility, so in that case it's better to just remove `VkSurfaceInfo`. Or just pass a null GD to `CreateSurface` from `VkSurfaceInfo` and skip validation in that case?

Jjagg:
> Or just pass a null GD to CreateSurface from VkSurfaceInfo and skip validation in that case?

Went ahead and did this one.

mellinoe:
@Jjagg Hey, I'm struggling a little bit to test this under Ubuntu 16.04. I've installed weston and tried to launch NeoDemo, but it fails to create a Vulkan Swapchain when I run with `SDL_VIDEODRIVER=wayland`. SDL2 also fails to create an OpenGL context ("Invalid Window").

> Under XWayland (when SDL acts as an X client, but the host DE uses Wayland), SDL returns Unknown for the subsytem in its SysWMinfo, so that doesn't work with the SwapchainSurface API

This seems to be what I'm hitting. Is there some step I'm missing?

Jjagg:
@mellinoe Are you running inside a weston shell? If you run `weston` in a terminal it will open a nested shell in which you can open a terminal to run the project.

mellinoe:
Yes, I'm running from a shell inside the weston compositor window that pops up.

> If this is what's happening it might be because of an old SDL version that doesn't support Wayland yet, is that possible? (I'm not sure if SDL would use X or crash in that case)

Hm, I did check that I was on the latest SDL2 package (2.0.8).

Jjagg:
I'm not sure what changed for me, but everything works for me. With SDL using either X or Wayland inside Weston I can create both OpenGL and Vulkan devices.

Jjagg:
I'm running the GettingStarted sample from veldrid-samples. Only changes are to make the reference to veldrid point to my local clone and create the graphicsdevice with Vulkan backend.

mellinoe:
I guess the other difference might be which drivers you are using. I have been using the Nvidia proprietary drivers, but I can try switching off them to check whether that makes a difference.

mellinoe:
Using the open source drivers, I get further but end up hitting an error when creating the Wayland Swapchain:

```
zxdg_surface_v6@22: error 3: xdg_surface has never been configured

Unhandled Exception: Veldrid.VeldridException: Unsuccessful VkResult: ErrorSurfaceLostKHR
```

Jjagg:
Damn, I don't really know what that's about.
I read in a similar bug report (but for glx, not vk) that it could be you need a newer Weston version or it might be a driver issue.

Jjagg:
Awesome :) I'm curious what went wrong with the other tests, but glad you got it running! No worries 😁

I also got this branch working with OpenWindow on Wayland btw. Still working on that when I can :)

PR #160, opened 2019-03-04
Fix typo in D3D11Swapchain

lweb20:
Change to "Height" instead of "Width"


mellinoe:
Looks good to me. Thanks for the fix!

PR #164, opened 2019-03-09
[D3D11, Vulkan] Handle 1D textures correctly when creating a TextureView

SomeAnon42:
The current code does not account for the existence of 1D textures.
- The Metal backend seems to handle this case correctly, though I can't verify that
- ~~OpenGL doesn't actually have TextureViews, so no changes are required~~
Wrong. There is in fact a code path that uses glTextureView, but it isn't taken when creating a full TextureView, which is why I'm not experiencing any issues. Changes might be required.
- D3D11 and Vulkan are covered by this PR

Maybe we should add some tests, though I'm not sure if it's possible to verify anything other than the fact that creating a TextureView succeeds for 1D textures.

### DIFF
--- a/src/Veldrid/D3D11/D3D11TextureView.cs
+++ b/src/Veldrid/D3D11/D3D11TextureView.cs
@@ -37,15 +37,33 @@ namespace Veldrid.D3D11
                 {
                     if (d3dTex.ArrayLayers == 1)
                     {
-                        uavDesc.Dimension = UnorderedAccessViewDimension.Texture2D;
-                        uavDesc.Texture2D.MipSlice = (int)description.BaseMipLevel;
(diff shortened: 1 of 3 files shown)


mellinoe:
Looks good to me; sorry for the delay in looking at this. I went ahead and added a test to the RenderTests class which creates a 1D texture and samples from it. OpenGL seems to be working without any changes.

mellinoe:
I did push a fix for OpenGL (it was binding things to the GL_TEXTURE_2D target rather than GL_TEXTURE_1D), although it didn't seem to cause any problems. I also ran the test on Metal, and everything is working fine there.

Thanks for catching this, everything should be resolved now.

PR #171, opened 2019-04-02
Add wrapper for SDL_SetRelativeMouseMode and SDL_CaptureMouse.

phelioz:
Add wrapper for SDL_SetRelativeMouseMode and SDL_CaptureMouse.
Also add comments for existing Sdl2.Mouse methods.

### DIFF
--- a/src/NeoDemo/Camera.cs
+++ b/src/NeoDemo/Camera.cs
@@ -22,22 +22,25 @@ namespace Veldrid.NeoDemo
         private float _yaw;
         private float _pitch;
 
-        private Vector2 _previousMousePos;
+        private Vector2 _mousePressedPos;
+        private bool _mousePressed = false;
         private GraphicsDevice _gd;
         private bool _useReverseDepth;
         private float _windowWidth;
         private float _windowHeight;
+        private Sdl2Window _window;
         private Sdl2ControllerTracker _controller;
 
         public event Action<Matrix4x4> ProjectionChanged;
         public event Action<Matrix4x4> ViewChanged;
 
-        public Camera(GraphicsDevice gd, float width, float height, Sdl2ControllerTracker controller)
+        public Camera(GraphicsDevice gd, Sdl2Window window, Sdl2ControllerTracker controller)
         {
             _gd = gd;
             _useReverseDepth = gd.IsDepthRangeZeroToOne;
-            _windowWidth = width;
-            _windowHeight = height;
+            _window = window;
+            _windowWidth = window.Width;
+            _windowHeight = window.Height;
             _controller = controller;
             UpdatePerspectiveMatrix();
             UpdateViewMatrix();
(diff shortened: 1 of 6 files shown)


mellinoe:
Sdl2Window actually already has a MouseDelta property which returns effectively the same information, and is updated whenever PumpEvents is called. See https://github.com/mellinoe/veldrid/pull/73 when it was added. Unfortunately, `InputSnapshot` is effectively frozen, because it is a public interface, and adding members to it is a breaking change. I believe that the "accumulation" pattern used by the current mouse delta tracker is also important -- multiple mouse events can be received in a single call to PumpEvents, so the deltas need to be accumulated (after being cleared right at the beginning of PumpEvents).

The SDL2 mouse changes look good to me.

phelioz:
Yeah right before the pull request I saw the MouseDelta inside Sdl2Window. The commit "Add wrapper for SDL_SetRelativeMouseMode and SDL_CaptureMouse" shouldn't have been a part of this pull request. It just slipped in by mistake, sorry!

mellinoe:
> The commit "Add wrapper for SDL_SetRelativeMouseMode and SDL_CaptureMouse" shouldn't have been a part of this pull request.

Did you mean that the other commit (with MouseDelta in it) shouldn't have been included? Right now there's only the SDL additions. Those look good to me, just wanted to make sure the PR is in the state you intended.

mellinoe:
> Shouldn't the line:
`Vector2 delta = new Vector2(mouseMotionEvent.x - _currentMouseX, mouseMotionEvent.y - _currentMouseY);`
Instead be:
`Vector2 delta = new Vector2(mouseMotionEvent.xrel, mouseMotionEvent.yrel);`

Possibly. I've actually experimented with the mouse capture functions in the past and had a really difficult time getting movement to work smoothly.

As part of this PR (or as a follow-up), I'd actually like to fix up NeoDemo so that when you click-drag the mouse around, the cursor stays in place (and potentially becomes invisible while dragging), similar to how most 3D editor views behave. Have you had any luck getting that kind of thing to work smoothly?

phelioz:
Yes I have actually exactly added that functionality to my own project, that's why I added SDL_SetRelativeMouseMode. 
SDL_SetRelativeMouseMode locks the mouse inside of the window and makes it invisible. So the position will not be updated and only xrel and yrel will be updated. You could then just enabled SDL_SetRelativeMouseMode on click and then disable it again on release.

phelioz:
Updated HandleMouseMotionEvent to use the relative mouse movement.
Also update NeoDemo so it now uses the SDL_SetRelativeMouseMode and uses the relative delta mouse movement.
Had to change InputTracker so I could access the Sdl2Window MouseDelta. Totally understand if you don't like those changes, just wanted to display how the camera could be fixed.

mellinoe:
@phelioz That is pretty nice as-is. The one additional thing I've tried to implement in the past is having the cursor "lock" to where you start the drag motion. Right now it is invisible while dragging, but it still moves and is in a different place when you stop dragging. Setting the cursor with `SDL_WarpMouseInWindow` (or `Sdl2Window.SetMousePosition` which uses the former) does work, but that also generates mouse motion events itself which need to be filtered somehow. Actually, I believe this was part of the reason that the xrel and yrel fields weren't being used. Have you had any luck getting something like that to work?

phelioz:
Yeah for just that use case SDL_SetRelativeMouseMode is not the best.
Added SDL_SetWindowGrab which that can used with SDL_WarpMouseInWindow to solve this problem.
Because with SDL_SetWindowGrab we can contain the mouse but control the mouse movement ourselves.
Then we calculated the the delta ourselves again and use that.

I think the mouse movement is smooth now, lowered the sensitivity a little bit also for smoother movement.

mellinoe:
@phelioz This is a great addition, thanks! Note that I added a commit on top to invert the Y-axis input in Camera.cs, because it was inverted compared to before.

The way you have implemented the mouse locking/hiding feels seamless and very smooth. It seems to be avoiding all of the problems that I was hitting with the approaches I've tried in the past. Thanks!

phelioz:
No problem, happy to help making Veldrid better :)

PR #175, opened 2019-04-11
Add more GL functions

RUSshy:


mellinoe:
@RUSshy As I mentioned in Gitter, I don't intend for Veldrid.OpenGLBindings to be a general-purpose OpenGL wrapper library that can be used on its own. Actually, I have long planned for these bindings to be moved into the core Veldrid library, and made private, in order to simplify packaging and releases. Because of that, I would really prefer to only maintain the minimal set of functions that Veldrid itself uses.

RUSshy:
Oh ok, i understand :)

PR #178, opened 2019-04-16
[D3D11] Fix SRV memory leak.

TheMostCuriousThing:


mellinoe:
Thanks for fixing this, the change looks good to me. Can you amend the commit author so that it matches your GitHub username? It will make things easier for me to keep track of.

TheMostCuriousThing:
Okay, should be good.

mellinoe:
LGTM. Thanks!

PR #180, opened 2019-04-24
[OpenGL] Set scissor rect relative to framebuffer, not viewport.

TheMostCuriousThing:
This fixes the misalignment of scissor rects when the viewport is not a full view of the framebuffer.


mellinoe:
@TheMostCuriousThing Thanks for catching this. The fix LGTM.

PR #194, opened 2019-06-19
Fix SamplerAddressMode.Clamp mapping in OpenGL

tzachshabtay:
The previous mapping to Clamp caused a glError of INVALID_ENUM on OSX, but ClampToEdge is supported.


mellinoe:
Thanks, according to info around the web, it seems like `GL_CLAMP` is definitely the wrong thing to be using here, although Nvidia doesn't seem to mind. Thanks for the fix!

PR #195, opened 2019-06-20
Fix mirror sampler address mode on metal

tzachshabtay:
Mapping Mirror to MirrorRepeat instead of MirrorClampToEdge to be consistent with the other backends (at the very least OpenGL backend which I tested it looks identical, but looking at the code I think the other backends as well).


mellinoe:
@tzachshabtay Thanks for the fix. This looks correct to me.

Issue #176, opened 2019-04-14
SamplerAddressMode axis mapping is different between backends


tzachshabtay:
I'm setting a custom SamplerDescription for my 2d game.
On OpenGL (on mac), addressModeU corresponds to the x-axis and addressModeV corresponds to the y-axis (like I would expect), but on Metal addressModeV corresponds to the x-axis and addressModeW corresponds to the y-axis.

mellinoe:
Ouch, you're correct about that. The [documentation](https://developer.apple.com/documentation/metal/mtlsamplerdescriptor) makes this clear. I'll push up a fix shortly.

mellinoe:
@tzachshabtay This should be fixed in the latest commit to master: https://github.com/mellinoe/veldrid/commit/be6d1db3d6d05344d35e0e0c57a8122a4305c395

tzachshabtay:
@mellinoe I am super confused currently. Now it seems that the OpenGL mappings have the same behavior I was previously describing metal had (i.e addressModeV corresponds to the x-axis and addressModeW corresponds to the y-axis). I could have sworn I tested this on OpenGL but maybe I didn't?

Anyway, looking at the ctor for SamplerDescription U/V/W correspond to R/S/T, and in OpenGL S&T correspond to X&Y, so I think actually that my assumption that addressModeU corresponds to the x-axis (which was based on the order of the arguments in SamplerDescription) was wrong and that the previous mapping for metal was right all along?
Sorry for confusing you...

PR #198, opened 2019-06-24
[Metal] Fix mappings for SamplerDescription.AddressModeU/V/W.

tzachshabtay:
Reverting commit: https://github.com/mellinoe/veldrid/commit/be6d1db3d6d05344d35e0e0c57a8122a4305c395

See: https://github.com/mellinoe/veldrid/issues/176#issuecomment-503817730

### DIFF
--- a/src/Veldrid/MTL/MTLSampler.cs
+++ b/src/Veldrid/MTL/MTLSampler.cs
@@ -18,9 +18,9 @@ namespace Veldrid.MTL
                 out MTLSamplerMipFilter mip);
 
             MTLSamplerDescriptor mtlDesc = MTLSamplerDescriptor.New();
-            mtlDesc.sAddressMode = MTLFormats.VdToMTLAddressMode(description.AddressModeU);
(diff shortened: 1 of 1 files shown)


mellinoe:
I think the correct fix is to actually replicate that commit (rather than revert it) for the OpenGL backend. The [documentation for MTLSamplerDescriptor](https://developer.apple.com/documentation/metal/mtlsamplerdescriptor?language=objc) is pretty clear that S/T/R correspond to width/height/depth. This is definitely the case for OpenGL as well, although it's hard to find a good reference for it. The mappings in OpenGLSampler.cs are currently incorrect, as they were for Metal.

tzachshabtay:
Ah, ok.
So the documentation in SamplerDescription probably needs to change as well.
For example:
https://github.com/mellinoe/veldrid/blob/80aa5db02df737c26f3e83e3ea7d662ac9892ce1/src/Veldrid/SamplerDescription.cs#L54

mellinoe:
Yes, those descriptions are also inaccurate.

mellinoe:
I made the change described above on the master branch, sot his shouldn't be needed. Let me know if you hit any issues after pulling that down.

PR #210, opened 2019-08-12
Cubemap from existing mip maps.

gleblebedev:
Need this to load existing mip levels from papermill specular cubemap:
https://github.com/ux3d/glTF-Sample-Environments/tree/master/papermill/specular


mellinoe:
This seems like a kind of unusual way to store the texture data, but the overload itself seems reasonable for accomplishing this. Thanks for the submission!

PR #214, opened 2019-08-27
"Not enough data in uniform buffer" error fixed.

gleblebedev:


mellinoe:
Thanks, not sure how this got mixed up.

PR #218, opened 2019-09-05
Bug in Veldrid.Utilities/BoundingBox Contains()

bmdub:
This one should be pretty self-explanatory; have a look...

Brian


mellinoe:
LGTM, thanks for the fix.

PR #219, opened 2019-09-07
Change ImGui update method to use protected methods

erictuvesson:
This allows for overriding the `Update` method and handle the input in a different way.

```cs
public new void Update(float deltaSeconds, object mycustominput)
{
    BeginUpdate(deltaSeconds);
    // handle custom input
    EndUpdate();
}
```

### DIFF
--- a/src/Veldrid.ImGui/ImGuiRenderer.cs
+++ b/src/Veldrid.ImGui/ImGuiRenderer.cs
@@ -372,14 +372,31 @@ namespace Veldrid
         /// </summary>
         public void Update(float deltaSeconds, InputSnapshot snapshot)
         {
+            BeginUpdate(deltaSeconds);
+            UpdateImGuiInput(snapshot);
+            EndUpdate();
+        }
+
+        /// <summary>
+        /// Called before we handle the input in <see cref="Update(float, InputSnapshot)"/>.
+        /// This render ImGui and update the state.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


mellinoe:
Can you give an example of how you intend to override this so I can get an idea of what it's needed for?

erictuvesson:
In my case I have a custom [input library](https://github.com/erictuvesson/Mallos.Input) so I will handle all the input in a different way than how Veldrid is handling it. I could create a full copy of the file and have that in my library instead, just it feels like having a lot of duplicate code when a small change can make it possible for be to handle it differently.

So all of this I would create my own version of.
https://github.com/mellinoe/veldrid/blob/9b324d3922c835cda52ee39b5f54ce86a07c3adf/src/Veldrid.ImGui/ImGuiRenderer.cs#L418-L481

This comes a little into the other PR's I have created, where I move out all of the input logic to make it possible to handle the base window without it. https://github.com/erictuvesson/veldrid/pull/1
It gives a little bit of abstraction and imo it makes it a little bit easier to read too.

mellinoe:
@erictuvesson Thanks for the explanation. That all sounds pretty reasonable to me. In general I am hesitant to make the ImGuiRenderer class a ton more flexible, because it can be tricky to get the ordering of native calls right, and allowing folks to re-order things and inject their own logic can be error-prone. On the other hand, I would expect that anyone creating their own Begin/Update/End loop here would know roughly what they are doing.

PR #220, opened 2019-09-07
Move key converter methods out from Sdl2Window class

erictuvesson:


mellinoe:
@erictuvesson What is the purpose of this change, exactly? The type is still internal, so it's not visible outside the assembly. Is this so you can source-include the file somewhere else and re-use it as an internal type?

erictuvesson:
The idea was just to make the Sdl2Window file more readable and not that big.

PR #221, opened 2019-09-07
Create IWindow for StartupUtilities

erictuvesson:
Let's depend less on `Sdl2Window` inside the utilities methods.


mellinoe:
`VeldridStartup` is intended to be a really basic utility class built on top of SDL2, but I don't think I want to make it a general "startup helper" for other systems to plug into. One important thing to note is that you can implement all of the "CreateGraphicsDevice" functions on top of Veldrid core without too much complicated logic. in this case, I'd rather encourage users to build on top of the `GraphicsDevice.CreateXYZ` family of functions directly.

PR #227, opened 2019-10-21
Fix off-by-one issue in Veldrid.Rectangle

csinkers:
Was doing some stuff with the Rectangle class defined in Veldrid.SDL2 and noticed it seems to have an off-by-one error. Compared against the behaviour of System.Drawing.Rectangle to verify.


mellinoe:
Good catch. Thanks for the submission!

PR #228, opened 2019-10-21
Add "position" and "size" Vector2 convenience properties to Veldrid.Rectangle.

csinkers:
Noticed that I was writing a lot of code like "new Vector2(rect.Width, rect.Height)" and "new Vector2(rect.X, rect.Y)" when working with rectangles, so thought it would be a good idea to add some properties to simplify things.


mellinoe:
Looks good to me. Thanks again!

PR #232, opened 2019-11-05
Allow Veldrid to run on D3D10-era hardware

SomeAnon42:
I don't think there's anything in Veldrid's D3D11 backend that prevents it from supporting older GPUs. Just tested my project on a laptop with an NVIDIA 9600M GT (which is a D3D10-era GPU), and this is all I had to change in Veldrid in order for it to work. 

Note that I had to remove the HResult check. That's because on my old laptop I'm getting a different error code when the debug runtime is missing (0x80004005, which can mean pretty much anything).

There's also one thing I'm wondering about. What is the point of trying to create a D3D device twice in a row with the exact same arguments? Could it be that the second call was supposed to have different arguments, and this is just a bug?

### DIFF
--- a/src/Veldrid/D3D11/D3D11GraphicsDevice.cs
+++ b/src/Veldrid/D3D11/D3D11GraphicsDevice.cs
@@ -59,23 +59,31 @@ namespace Veldrid.D3D11
                 flags = DeviceCreationFlags.Debug;
             }
 
+            SharpDX.Direct3D.FeatureLevel[] featureLevels = new[]
+            {
+                SharpDX.Direct3D.FeatureLevel.Level_11_1,
+                SharpDX.Direct3D.FeatureLevel.Level_11_0,
+                SharpDX.Direct3D.FeatureLevel.Level_10_1,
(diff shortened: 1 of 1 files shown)


mellinoe:
I don't have a problem with allowing lower-end devices to work, as long as there's not cascading effects/checks throughout the codebase. This looks reasonable if it's all that is involved. However, I suspect that some of these values should be different if you aren't running on D3D11 hardware:

https://github.com/mellinoe/veldrid/blob/master/src/Veldrid/D3D11/D3D11GraphicsDevice.cs#L90-L108

> There's also one thing I'm wondering about. What is the point of trying to create a D3D device twice in a row with the exact same arguments? Could it be that the second call was supposed to have different arguments, and this is just a bug?

That looks like a bug to me.

SomeAnon42:
Alright then, looks like I will have to do some more research and testing.

SomeAnonDev:
Apologies for taking so long. I plan to get back to this PR once (if?) #285 is merged.

PR #236, opened 2019-11-20
Add event to optionally abort closing the SDL2 window

swoolcock:
Subscribing to `CloseRequested` and returning `true` will abort the close action.  This allows users to provide "are you sure?" dialogs, etc.

This will catch both programmatic `Close()` calls and explicit user actions (clicking the window close button, Alt+F4, Cmd+Q).

Tested on macOS by manually inserting into NeoDemo: `_window.CloseRequested += () => true;`

### DIFF
--- a/src/Veldrid.SDL2/Sdl2Window.cs
+++ b/src/Veldrid.SDL2/Sdl2Window.cs
@@ -43,6 +43,7 @@ namespace Veldrid.Sdl2
         private string _cachedWindowTitle;
         private bool _newWindowTitleReceived;
         private bool _firstMouseEvent = true;
+        private Func<bool> _closeRequestedHandler;
 
         public Sdl2Window(string title, int x, int y, int width, int height, SDL_WindowFlags flags, bool threadedProcessing)
         {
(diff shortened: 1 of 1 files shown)


mellinoe:
I have wanted this kind of functionality in the past, so this is a good submission. Thanks!

The one thing I find confusing about this is that CloseRequested is an `event`. This naturally means that you can subscribe many callbacks to it, but that does not make a lot of sense in this case. All of the callbacks will be invoked, but only the last will affect whether the window remains open, which makes it a bit misleading. I'd prefer if the API were exposed in a way that made it obvious that only one callback could be active. As an example:

```C#
public class Sdl2Window
{
    public void SetCloseRequestedHandler(Func<bool> handler) { closeRequestedHandler_ = handler; }
}
```

swoolcock:
Will do.

swoolcock:
@mellinoe are these changes ok?  Would love to get this in so we can use it on our side. 😅

mellinoe:
@swoolcock Yes, looks great now. Thanks for making the requested change, and sorry for the delay.

PR #237, opened 2019-11-20
Add support for Windows/Command keys

swoolcock:
Called "Gui" keys in SDL2 scancodes and keymods, so I have mapped the scancodes to the existing `Key.LWin` and `Key.RWin` cases and created a new `Gui` case in `ModifierKeys`.

### DIFF
--- a/src/Veldrid.SDL2/ModifierKeys.cs
+++ b/src/Veldrid.SDL2/ModifierKeys.cs
@@ -7,5 +7,6 @@
         Alt = 1,
         Control = 2,
         Shift = 4,
+        Gui = 8,
     }
 }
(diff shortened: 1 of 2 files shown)


mellinoe:
This is a good addition, but I'm not sure about the name "Gui". Although it sounds OS-specific, I would be fine just calling it "Win" or "Windows" -- the purpose is immediately obvious. Some Linux operating systems call the key "Meta" or "Super", but I don't think those are widely recognized. Is "Gui" a recognized term for this?

swoolcock:
I used "Gui" because that's what the scancodes are called in SDL2, however I have seen the modifier called "Super" in OpenTK.  Should we rename it to that?

mellinoe:
Sorry for the delay on this. I looked around, and found even more alternative names for this key. So there's definitely no consensus. Since this is in the SDL2 package and directly wraps SDL2 keys, I'm just going to merge this in with the original name `Gui`. Thanks for the patience.

mellinoe:
Merged manually. Thanks again.

PR #241, opened 2019-12-18
Custom OpenGL TextureTarget

gleblebedev:
Custom OpenGL TextureTarget.

This is required to support platform specific targets like external texture on Android.

### DIFF
--- a/src/Veldrid/BackendInfoOpenGL.cs
+++ b/src/Veldrid/BackendInfoOpenGL.cs
@@ -1,5 +1,4 @@
-﻿
-#if !EXCLUDE_OPENGL_BACKEND
+﻿#if !EXCLUDE_OPENGL_BACKEND
 using System;
 using Veldrid.OpenGL;
 using Veldrid.OpenGLBinding;
--- a/src/Veldrid/BackendInfoOpenGL.cs
+++ b/src/Veldrid/BackendInfoOpenGL.cs
@@ -1,8 +1,8 @@
 ﻿
-using Veldrid.OpenGLBinding;
 #if !EXCLUDE_OPENGL_BACKEND
 using System;
 using Veldrid.OpenGL;
+using Veldrid.OpenGLBinding;
 
 namespace Veldrid
 {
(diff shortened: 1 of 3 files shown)


src/Veldrid/BackendInfoOpenGL.cs
```diff
@@ -33,6 +34,13 @@ internal BackendInfoOpenGL(OpenGLGraphicsDevice gd)
         /// </summary>
         /// <returns>The Veldrid Texture's underlying OpenGL texture name.</returns>
         public uint GetTextureName(Texture texture) => Util.AssertSubtype<Texture, OpenGLTexture>(texture).Texture;
+
+        /// <summary>
+        /// Sets the texture target of the OpenGL texture object wrapped by the given Veldrid Texture to to a custom value.
+        /// This could be used to set platform specific texture target values like Veldrid.OpenGLBinding.TextureTarget.TextureExternalOes.
+        /// </summary>
+        public void SetTextureTarget(Texture texture, uint textureTarget) => Util.AssertSubtype<Texture, OpenGLTexture>(texture).TextureTarget = (TextureTarget)textureTarget;
```
mellinoe: Is it possible for the OpenGL constructor to query the texture for its target? Apparently it's possible directly in OpenGL 4.5, but I don't see anything in GLES where we need it. Others mention just trying to bind the texture to various targets and seeing which one works. If there's a way we can avoid having this extra function, that would be great. Otherwise, this seems like a reasonable workaround.

gleblebedev: I don't think it is possible to query it as one can create texture name without binding it to any target. Also it seems that OpenGL allows to successfully bind texture name to different targets so there is no way to query it by try :(

mellinoe: Ok, thanks for investigating. I'm not terribly concerned about polluting the BackendInfo* types since they are fairly advanced and niche.

mellinoe:
Merged manually. Thanks for the contribution and for digging into the internals to make this scenario work.

PR #242, opened 2019-12-19
Create D3D11 device with adapter and flags

gleblebedev:


mellinoe:
I think this is a good addition. No reason we can't support this kind of thing. In fact, it's very similar to one of the [`CreateVulkan` overloads](https://github.com/mellinoe/veldrid/blob/master/src/Veldrid/GraphicsDevice.cs#L961) that exists already. I'd ask that you make this an extension of that pattern, and add a `D3D11DeviceOptions` structure, similar to [VulkanDeviceOptions](https://github.com/mellinoe/veldrid/blob/master/src/Veldrid/VulkanDeviceOptions.cs).

gleblebedev:
@mellinoe done.

mellinoe:
I believe you have forgotten to add the source for D3D11DeviceOptions in the latest version.

I'll add one extra request: please change all instances of `SharpDX.Direct3D11.DeviceCreationFlags` to `uint` in the public API. They can be represented however you want internally in Veldrid, but we should be careful to not expose any more SharpDX types in the interface.

gleblebedev:
Sorry for delay. Missing file added. The flags were already changed to uint.

src/Veldrid/D3D11/D3D11Texture.cs
```diff
@@ -143,6 +143,12 @@ public D3D11Texture(Device device, ref TextureDescription description)
             }
         }
 
+        public D3D11Texture(Texture2D existingTexture, TextureType type):
```
mellinoe: Can you explain what some of these additional Texture-related changes are for?

src/Veldrid/D3D11/D3D11GraphicsDevice.cs
```diff
@@ -113,6 +130,31 @@ public D3D11GraphicsDevice(GraphicsDeviceOptions options, SwapchainDescription?
             PostDeviceCreated();
         }
 
+        private static bool SdkLayersAvailable()
```
mellinoe: Can you cache the result of this function so that the initial evaluation only needs to happen once? It involves creating a whole dummy Device, so it's fairly expensive.

src/Veldrid/D3D11/D3D11Texture.cs
gleblebedev: In certain citation I have to create a texture provided from external source. I've used this to create a texture with correct format.

I can actually roll back this as o have to move the format resolution code out of main Veldrid repository.

gleblebedev: @mellinoe fixed

src/Veldrid/D3D11/D3D11GraphicsDevice.cs
gleblebedev: @mellinoe  fixed

mellinoe:
LGTM. Thanks for the submission and follow-up fixes.

PR #244, opened 2020-01-02
Added delegates for SDL_GetDesktopDisplayMode and SDL_GetNumVideoDisplays

drjaydenm:
I needed to use these two functions from SDL and couldn't see them in the Sdl2Native class so just thought I would add them in.
SDL_GetDesktopDisplayMode is useful to get the desktop resolution in a cross-platform way.


mellinoe:
LGTM, thanks for the submission!

Issue #247, opened 2020-01-14
Render target binding as texture view D3D11


drjaydenm:
Hi @mellinoe, this is the bug that I was talking about on Gitter the other day where using a render target as a texture view binding doesn't always work on D3D11 backend.

Tl;dr of the bug as far as I can see is:
* First render pass writes to render target 1 and has render target 2 as a texture view binding
* A second render pass runs writing to render target 2 with a different shader which has no texture bindings (this implicitly unbinds render target 2 from the first shader)
* A third render pass runs writing back to render target 1 with render target 2 as a texture view binding again - this time, the texture view binding doesn't get bound back to the shader properly

I've done a little digging and it looks like when render target 2 gets unbound, it isn't removed from the list of bound texture views in D3D11CommandList, therefore in the third pass, it thinks it is already bound and doesn't actually bind it properly (I could be way off here though).
Let me know if you have any questions

PR for the repro test is here: #248

mellinoe:
Thank you for the detailed description and for the repro code. The problem is very likely a bug in the tracking of bound resources, esp. regarding render target and shader resource views. I will take a closer look at this soon.

mellinoe:
This should be fixed now: https://github.com/mellinoe/veldrid/commit/9b16d7e9390b8cf466c4b21232313418f0e8b2c7

drjaydenm:
Thanks for taking a look and getting a fix up so quickly 😄

PR #248, opened 2020-01-14
Add texture binding test

drjaydenm:
Hi @mellinoe,

I've added in a RenderTest as you suggested which can reproduce the bug on the D3D11 backend. I've also given you access to modify the branch so feel free to change whatever you feel like, no need to check with me first.

I loaded the details into the issue here: #247


mellinoe:
Thanks for submitting this working test. I've merged it along with a fix which allows the test to pass.

PR #252, opened 2020-02-11
[OpenGL] Reset the draw buffers after clearing a color target.

### LINKED ISSUE #251: [OpenGL] [Bug] ClearColorTarget doesn't reset the color attachments
`OpenGLCommandExecutor.ClearColorTarget` calls `glDrawBuffers` to limit which colour target gets cleared, but doesn't reset it afterwards, which makes it so only that target gets rendered to.

From some simple testing, it seems this bug can be fixed by simply adding this snippet of code to the end of that function:
```CSharp
    if (!_isSwapchainFB)
    {
        int colorCount = _fb.ColorTargets.Count;
        DrawBuffersEnum* bufs = stackalloc DrawBuffersEnum[colorCount];
        for (int i = 0; i < colorCount; i++)
        {
            bufs[i] = DrawBuffersEnum.ColorAttachment0 + i;
        }
        glDrawBuffers((uint)colorCount, bufs);
        CheckLastError();
    }
```

Doom2fan:
This fixes `OpenGLCommandExecutor.ClearColorTarget` not resetting the draw buffers after clearing the target.
(Fixes #251)

### DIFF
--- a/src/Veldrid/OpenGL/OpenGLCommandExecutor.cs
+++ b/src/Veldrid/OpenGL/OpenGLCommandExecutor.cs
@@ -77,6 +77,18 @@ namespace Veldrid.OpenGL
             {
                 glEnable(EnableCap.ScissorTest);
             }
+
+            if (!_isSwapchainFB)
+            {
+                int colorCount = _fb.ColorTargets.Count;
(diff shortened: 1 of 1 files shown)


mellinoe:
LGTM, thanks for the fix!

PR #253, opened 2020-02-11
[Vulkan] Replace some foreachs that were causing lots of allocations.

Doom2fan:
These instances of `foreach` were causing a lot of allocations, making the GC trigger extremely often.

### DIFF
--- a/src/Veldrid/Vk/VkCommandList.cs
+++ b/src/Veldrid/Vk/VkCommandList.cs
@@ -319,9 +319,9 @@ namespace Veldrid.Vk
 
                     // Increment ref count on first use of a set.
                     _currentStagingInfo.Resources.Add(vkSet.RefCount);
-                    foreach (ResourceRefCount refCount in vkSet.RefCounts)
+                    for (int i = 0; i < vkSet.RefCounts.Count; i++)
                     {
-                        _currentStagingInfo.Resources.Add(refCount);
+                        _currentStagingInfo.Resources.Add(vkSet.RefCounts[i]);
                     }
                 }
 
--- a/src/Veldrid/Vk/VkFramebuffer.cs
+++ b/src/Veldrid/Vk/VkFramebuffer.cs
@@ -239,8 +239,9 @@ namespace Veldrid.Vk
 
         public override void TransitionToIntermediateLayout(VkCommandBuffer cb)
         {
-            foreach (FramebufferAttachment ca in ColorTargets)
+            for (int i = 0; i < ColorTargets.Count; i++)
             {
+                FramebufferAttachment ca = ColorTargets[i];
                 VkTexture vkTex = Util.AssertSubtype<Texture, VkTexture>(ca.Target);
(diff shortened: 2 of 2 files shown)


peppy:
The cases you have changed are simple `List` / `ReadOnlyList` usages with no custom enumerators, which have been optimised away for a good time now. Very curious what runtime you are running on to see such allocations. Are you able to provide `BenchmarkDotNet` `[MemoryDiagnoser]` results showing the difference?

Doom2fan:
> 
> 
> The cases you have changed are simple `List` / `ReadOnlyList` usages with no custom enumerators, which have been optimised away for a good time now. Very curious what runtime you are running on to see such allocations. Are you able to provide `BenchmarkDotNet` `[MemoryDiagnoser]` results showing the difference?

I couldn't get `BenchmarkDotNet` to work, and I'm not sure how I'd do tests with it for such a thing anyway, so here's the results from VS' ".NET Object Allocation Tracking" profiling option.

With `foreach`:
4x5ez65
With `for`:
Untitled-1

Both were made on my engine running with .NET Core 3.1, "Release" configuration and x64 platform. As you can see, there's a ton of allocations going on there with the `foreach` version.

Additionally, `VkSwapchainFramebuffer` also implements the two functions I changed in `VkFramebuffer`, and its versions of them already used a `for` loop for iterating `ColorTargets`, so I believe `foreach` being used here was accidental.

peppy:
Aha, this is the classic case of accessing via `IReadOnlyList` adding enumerator overhead.

```csharp
using System;
using System.Collections.Generic;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;

namespace EnumeratorAllocs
{
    [MemoryDiagnoser]
    public class Program
    {
        static void Main(string[] args)
        {
            BenchmarkRunner.Run(typeof(Program));
        }

        private List<object> list = new List<object>();

        private IReadOnlyList<object> readOnlyList => list;

        [GlobalSetup]
        public void SetUp()
        {
            for (int i = 0; i < 10000; i++)
                list.Add(new object());
        }

        [Benchmark]
        public void EnumerateOverList()
        {
            foreach (var o in list)
            {

            }
        }
        
        [Benchmark]
        public void EnumerateOverReadOnlyList()
        {
            foreach (var o in readOnlyList)
            {

            }
        }
    }
}
```

```

|                    Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
|         EnumerateOverList | 30.14 us | 0.154 us | 0.136 us |     - |     - |     - |         - |
| EnumerateOverReadOnlyList | 88.53 us | 0.861 us | 0.805 us |     - |     - |     - |      40 B |
```

mellinoe:
Yes, this is a classic problem. I've done my best to squash this pattern everywhere else in the library, but it has crept back into some more recent additions. Thanks for the fix!

RUSshy:
It is easy to see what is the issue, structs needs IEquatable<T> and overriding Equals, so it avoid boxing allocations most most cases

PR #256, opened 2020-03-05
[D3D11] Replace a foreach that was causing a lot of allocations.

Doom2fan:
Pretty much the same issue Vulkan had, trivial fix.


mellinoe:
LGTM, thanks for the fix.

PR #260, opened 2020-03-09
[OpenGL] Fix the stencil ops not being set.

Doom2fan:


mellinoe:
Thanks, this also LGTM. It would be great to get some tests that validate stencil functionality is working, or at least some debug piece in NeoDemo that can be manually checked.

PR #262, opened 2020-03-20
Added a "IsDisposed" property to some public classes.

Doom2fan:
A pretty useful thing to have. Having a property like this allows some automatic clean-up to be done, which cannot be done otherwise, since there's no way to identify if the classes have been disposed of and are now useless.

### DIFF
--- a/src/Veldrid/Buffer.cs
+++ b/src/Veldrid/Buffer.cs
@@ -26,6 +26,11 @@ namespace Veldrid
         public abstract string Name { get; set; }
 
         /// <summary>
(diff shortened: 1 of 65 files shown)


mellinoe:
I manually rebased and merged this into master, along with some disposal tests that validate the behavior is the same across backends. Along with that, I changed the IsDisposed implementation for most OpenGL types to return `true` immediately when Dispose is called, instead of when the resource disposal actually happens (because it is deferred in the OpenGL backend).

Thanks again for the contribution!

PR #263, opened 2020-03-23
[OpenGL] Fixed cubemap framebuffers.

Doom2fan:


src/Veldrid/OpenGL/OpenGLFramebuffer.cs
```diff
@@ -77,9 +104,9 @@ public void CreateGLResources()
                         glFramebufferTextureLayer(
                             FramebufferTarget.Framebuffer,
                             GLFramebufferAttachment.ColorAttachment0 + i,
-                            glTex.Texture,
+                            (uint)textureTarget,
                             (int)colorAttachment.MipLevel,
-                            (int)colorAttachment.ArrayLayer);
+                            (int)colorAttachment.ArrayLayer / 6);
```
mellinoe: This portion isn't quite correct -- `glFramebufferTextureLayer` expects slightly different parameters and the texture name itself is supposed to be passed here. Also, the array layer shouldn't be divided by 6 -- the face index is included in that layer.

mellinoe:
I branched off of this and made quite a few additional cubemap-related fixes. I've also added much more substantial test coverage for rendering to cubemap faces via Framebuffers, and tested it on all backends. Several other fixes were needed in OpenGL and other backends to get it all working correctly. Thanks for getting this started!

Closing since this was merged manually.

PR #269, opened 2020-04-12
[OpenGL] Attempted to fix depth targets in cubemap framebuffers.

Doom2fan:


mellinoe:
Looks good, thanks for the fix!

PR #270, opened 2020-04-13
fixed for layerless cubemap binding

geoeo:
This is a fix for what Doom2Fan tried to do with his pull request. We needed the  glFramebufferTexture https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glFramebufferTexture.xhtml, without a layer. Otherwise only one cubemap layer was writable as a framebuffer target. Im not sure about his code, it might be redundant now. But I left it in


Doom2fan:
Like I said in the Gitter chat, this isn't the right way to do it - this will change the current behaviour of cubemaps from it being possible to draw to a single face to always having the full cubemap bound as a layered target.
I believe the correct way to do this is by adding an option for binding an entire cubemap or texture array as a layered target.

geoeo:
Doom2Fan is working on a proper solution

PR #271, opened 2020-04-14
Add support for layered framebuffers

Doom2fan:
Layered framebuffers are useful as they can be used to render geometry to things like cubemaps in a single drawcall through geometry shaders.
Originally I was going to add an option to let the user specify how many layers to bind instead, but it appears OpenGL doesn't support this, so I had to scrap that and bind all layers instead.

Unfortunately, actually drawing anything is untested, as I have no geometry shaders to test it with, and Veldrid.Spirv can't compile geometry shaders anyway.
But at least OpenGL *should* work, as I based that on PR #270, and that supposedly works.

Issues:
* OpenGL won't clear the stencils for some reason. (Might be all stencil writing that won't work? Couldn't test, see last one)
* No Metal implementation. (Don't know Metal and can't test)
* Couldn't test drawing to all layers due to lack of geometry shaders.

### DIFF
--- a/src/Veldrid.OpenGLBindings/OpenGLNative.cs
+++ b/src/Veldrid.OpenGLBindings/OpenGLNative.cs
@@ -870,6 +870,19 @@ namespace Veldrid.OpenGLBinding
             => p_glBlitFramebuffer(srcX0, srcY0, srcX1, srcY1, dstX0, dstY0, dstX1, dstY1, mask, filter);
 
         [UnmanagedFunctionPointer(CallConv)]
+        private delegate void glFramebufferTexture_t(
+            FramebufferTarget target,
+            GLFramebufferAttachment attachment,
+            uint texture,
+            int level);
+        private static glFramebufferTexture_t p_glFramebufferTexture;
+        public static void glFramebufferTexture(
+            FramebufferTarget target,
+            GLFramebufferAttachment attachment,
+            uint texture,
+            int level) => p_glFramebufferTexture(target, attachment, texture, level);
+
+        [UnmanagedFunctionPointer(CallConv)]
         private delegate void glFramebufferTextureLayer_t(
             FramebufferTarget target,
             GLFramebufferAttachment attachment,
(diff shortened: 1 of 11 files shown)


geoeo:
I will check this out tomorrow. I can let you know if the OpenGL part works

mellinoe:
A few things, briefly:

* Since this won't be used immediately in NeoDemo or a sample, we need to add a test for it. It can be as simple as rendering a different solid color to each layer, but there should be some validation on that level, at least. The existing RenderTests are a good start, it doesn't need to be much more complicated than those.
* Even if Veldrid.SPIRV doesn't support geometry shaders, we can still hand-write some GLSL, HLSL, and/or SPIR-V shaders and feed them to Veldrid.
* From what I can see, it's also possible to render to different layers of a layered Framebuffer from a vertex shader, but it's not universally supported:
  * If the `GL_ARB_shader_viewport_layer_array` GLSL extension is enabled, then `gl_Layer` can be written to in a vertex shader.
  * If [this D3D11 flag](https://docs.microsoft.com/en-us/windows/win32/api/d3d11/ns-d3d11-d3d11_feature_data_d3d11_options3) is set, then you can write to the `SV_RenderTargetArrayIndex` semantic from the vertex stage.
  * If this is feasible, I'd much prefer a test that uses a vertex shader to target different layers -- it will be more portable.
* It needs to be supported on Metal as well. [Metal does have this functionality](https://developer.apple.com/documentation/metal/mtlrenderpassdescriptor/rendering_to_multiple_texture_slices_in_a_draw_command?language=objc). It would likely need to be tested with a vertex shader targeting different layers, as mentioned above.

Given the weirdness of geometry shaders and the lack of universal support for them, I'd feel a lot more comfortable with this addition if I could see that it's fully usable with just vertex/fragment shaders.

mellinoe:
Oops, I did not mean to close this.

mellinoe:
Another note: if support for rendering to individual layers of a Framebuffer is conditional on some extensions (OpenGL) or feature levels (D3D11) being present, we should add those as flags on `GraphicsDeviceFeatures` so it's clear when they can be used.

geoeo:
Are mip levels implicitly created for a depth cube map? Because in my LayoutDescription I set 1 for mip map levels, but RenderDoc shows 6.

geoeo:
Seems to work for OpenGL. But as @mellinoe said making a test explicitly for this would be best. I also cant speak for mip mapping. I tested setting the texture layout description to >1 and in RenderDoc the other mip levels were black. But maybe im missing something. Just wanted to let you know

geoeo:
P.S. I just want to mention it. Sampling cube map textures seems to be flipped for me in OpenGL. This might be a bug in my code. But I just want to mention it.

geoeo:
Is there any branch where this is being developed? Im still having this sampling issue. In renderdoc the cubemap looks fine, but when I sample it, I seem to be sampling wrong texels in the cubemap. I would like to see what/if anything fixes this. Thank you!

geoeo:
There is definitely a regression for opengl cubemaps. It seems when binding a layered texture as cubemap the faces are not set correctly. E.g. an older skybox shader I have has faces flipped. 

This is the "new" indexing that seems to give correct sampler results. But I still have to negate x and y values in the shader so its a bit messed up.

    
            Right = 1, // +X
            Left = 0, // -X
            Top = 3, // +Y
            Bottom = 2, // -Y
            Back = 4, // +Z
            Front = 5 // -Z

PR #281, opened 2020-06-03
Add key repeat info to KeyEvent

trolleyman:
Small change to add key repeat info to KeyEvent

Also - quick question - what's the policy on breakages for Veldrid? If I didn't have the default parameter for `repeat` on the `KeyEvent` constructor, would that be a breaking change? If so, would you have to release a new major version? i.e. `5.0.0`, or would it be a minor version increment? i.e. `4.8.0`

Also - is the change of the `ToString` function breaking behaviour, in your opinion?

### DIFF
--- a/src/Veldrid.SDL2/KeyEvent.cs
+++ b/src/Veldrid.SDL2/KeyEvent.cs
@@ -5,14 +5,16 @@
         public Key Key { get; }
         public bool Down { get; }
         public ModifierKeys Modifiers { get; }
-        public KeyEvent(Key key, bool down, ModifierKeys modifiers)
+        public bool Repeat { get; }
+        public KeyEvent(Key key, bool down, ModifierKeys modifiers, bool repeat = false)
         {
             Key = key;
             Down = down;
             Modifiers = modifiers;
+            Repeat = repeat;
         }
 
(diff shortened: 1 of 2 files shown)


mellinoe:
Thanks for the contribution. This look good overall.

Regarding breaking changes and major versions: the policy is to not make any breaking changes. Bumping the major version to 5.0.0 would only be done for a major restructuring of the library.

>  If I didn't have the default parameter for repeat on the KeyEvent constructor, would that be a breaking change?

Actually, this change as-is would be considered a breaking change, because it effectively removes the 3-argument constructor and adds a 4-argument constructor. It needs to be adjusted so that there is both a 3- and 4-argument constructor. The 3-argument constructor would behave as if `false` was passed in for the new property.

mellinoe:
@trolleyman I merged this in with one additional commit on top:

https://github.com/mellinoe/veldrid/commit/a8001529b6b06be9b49e46a72d5ac2d0e24b6f21

Thanks again for the contribution!

trolleyman:
Ah right, thanks for that catch

No problem - thanks for merging this :)

Just a heads up - will probably be doing a few more of these in the next few weeks, as we're looking into using Veldrid at work

PR #282, opened 2020-06-03
Add SDL_GetModState

trolleyman:
This allows users to check if a lock key is locked - there is no way of telling from keyboard events alone

Potentially this could also be expanded to allow every modifier key to be checked, and to provide only a small wrapper around GetModState, if this isn't acceptable

### DIFF
--- a/src/Veldrid.SDL2/LockKey.cs
+++ b/src/Veldrid.SDL2/LockKey.cs
@@ -1,8 +1,0 @@
-﻿namespace Veldrid
-{
-    public enum LockKey
-    {
-        CapsLock,
-        NumLock,
-    }
-}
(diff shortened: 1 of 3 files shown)


mellinoe:
Exposing SDL_GetModState is a good addition. I'm less sure about adding Veldrid.LockKey and Sdl2Window.IsKeyLocked. The latter two seem useful, but aren't logically tied to a window. They are global state queries. I'd rather leave those additions out and just have users call SDL_GetModState directly (or wrap it in something like NeoDemo's InputManager class).

trolleyman:
Yeah that's a fair point - I was a bit iffy about it myself. I've removed that stuff from the MR, and left just the SDL_GetModState addition

mellinoe:
LGTM. Thanks for the update.

PR #283, opened 2020-06-03
Update veldrid.imagesharp code to use image sharp rc1

UkooLabs:
Updated code to use latest rc1 of imagesharp

### DIFF
--- a/src/Veldrid.ImageSharp/ImageSharpCubemapTexture.cs
+++ b/src/Veldrid.ImageSharp/ImageSharpCubemapTexture.cs
@@ -169,13 +169,38 @@ namespace Veldrid.ImageSharp
                         Format,
                         TextureUsage.Sampled | TextureUsage.Cubemap));
 
-            for (int level = 0; level < MipLevels; level++){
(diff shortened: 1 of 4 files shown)


feliwir:
Would appreciate this getting merged quickly, since we're stuck with ImageSharp beta 6 currently when using Veldrid (last major release was on beta 6)

mjkkirschner:
Isn't this version of imageSharp the first one to change the license to AGPL - which may complicate treating this repo as MIT licensed?

https://github.com/SixLabors/ImageSharp/blob/master/LICENSE

Or maybe this happened already in one of the previous beta versions?

UkooLabs:
@mjkkirschner It should be fine as SixLabors said If you had an existing open source repo that used ImsgeSharp before it became RC you can continue to use InsgeSharp

JimBobSquarePants:
@UkooLabs is correct. The readme states.

> Open Source projects who have taken a dependency on ImageSharp prior to adoption of the AGPL v3 license are permitted to use ImageSharp (including all future versions) under the previous Apache 2.0 License.

JimBobSquarePants:
And like that... any ambiguity goes away 😄

PR #284, opened 2020-06-11
Make SDL_ControllerDeviceEvent public

trolleyman:
Seems like an oversight - I'm going to use this event to handle the ControllerDeviceAdded/Removed/Remapped events https://wiki.libsdl.org/SDL_ControllerDeviceEvent


mellinoe:
LGTM, thanks for the fix

PR #288, opened 2020-07-06
DCI-P3 support for Retina Display

OpenPachinko:
On MacOS, DCI-P3 color space and 1 billion color display are enabled in Retinal Display by specifying R16_G16_B16_A16_Float.

### DIFF
--- a/src/Veldrid/MTL/MTLSwapchain.cs
+++ b/src/Veldrid/MTL/MTLSwapchain.cs
@@ -83,7 +83,7 @@ namespace Veldrid.MTL
             }
 
             PixelFormat format = description.ColorSrgb
-                ? PixelFormat.R16_G16_B16_A16_Float;
+                ? PixelFormat.R16_G16_B16_A16_Float; // Deep Color (DCI-P3 support for Retina Display)
                 : PixelFormat.B8_G8_R8_A8_UNorm;
 
             _metalLayer.device = _gd.Device;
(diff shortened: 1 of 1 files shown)


mellinoe:
I don't know enough about the impact this will have for folks on different types of hardware. Is this supported on non-Retina displays, for example? The documentation for the property is not very informative, but it suggests that RGBA16Float format should be supported everywhere regardless of hardware. If anyone can test this on an older mac of some kind, that would be very helpful.

Other than that, the change seems reasonable, but I don't know if this is the best way to expose it. In reality, there are a number of valid swapchain formats that are available, and Veldrid hides that from you. I think a better solution (and more "portable" in a sense) would be to expose a `PixelFormat ColorFormat` field on `SwapchainDescription`, akin to the existing `DepthFormat` field. This would supercede the `ColorSrgb` field if specified. This is preferable because other backends (Vulkan and D3D11, mainly) also support a wider variety of Swapchain formats and could also be controlled with this field.

One last thing: VeldridStartup.CreateMetalGraphicsDevice is currently not passing along the SRGB flag correctly, so this change wasn't having an effect through that codepath (although it should have been working if you created a Metal GraphicsDevice some other way. I'll fix that in the meantime.

OpenPachinko:
古いThunderbolt DisplayとMacBook Airの2種類で試しました。
Non-Retina Displayでも問題はありません。
macOSは適切な形式で出力するようです。

OpenPachinko:
I tried it on an old Thunderbolt Display and a MacBook Air.
Non-Retina Display is no problem.
MacOS seems to output in the proper format.

PR #299, opened 2020-08-16
Fixed .netcore compile

onehundredfeet:
Added namespace to ambiguous classes - NativeLoader (Used Veldrid one)

Tiny change. Should be ultra-low risk.

### DIFF
--- a/src/NeoDemo/ImGuiRenderable.cs
+++ b/src/NeoDemo/ImGuiRenderable.cs
@@ -11,6 +11,8 @@ namespace Veldrid.NeoDemo
         private int _width;
         private int _height;
 
+        public ImGuiRenderer Renderer => _imguiRenderer;
+
         public ImGuiRenderable(int width, int height)
         {
             _width = width;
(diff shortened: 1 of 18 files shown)


mellinoe:
I don't understand the reasoning behind these changes, they don't seem necessary as these projects don't have any ambiguous reference problems with their current compilation target. Also there are quite a few unrelated changes and style adjustments that have been mixed in as well. Feel free to reopen if there's something that should be pulled out of here.

onehundredfeet:
I should have canceled the request. Thanks for closing. My git knowledge is incomplete and the pull request seemed to keep including further updates rather than locking at the change I submitted. 

Sorry for the confusion.

onehundredfeet:
With .netcore there is a ambiguous reference.  NativeLibrary class which was causing conflicts with your code. The tiny change made it compile. You can ignore if you want, but it was helpful for me to have it compiling natively with .net core.

Ignore the merges. I think that was my poor git knowledge at the time. The only thing to look at were the changes in the first submission.

PR #300, opened 2020-08-19
Add support for float64 in shaders

### LINKED ISSUE #296: How do I enable VkPhysicalDeviceFeatures::shaderFloat64?
An exception was thrown when creating a graphics pipeline with its vertex shader using `dvec2`:
```
Unhandled exception. Veldrid.VeldridException: A Vulkan validation error was encountered: [ErrorEXT] (UnknownEXT)  [ UNASSIGNED-CoreValidation-Shader-FeatureNotEnabled ] Object: VK_NULL_HANDLE (Type = 0) | Shader requires VkPhysicalDeviceFeature
s::shaderFloat64 but is not enabled on the device
```
I do not fully understand what that means, but physical device features seem [hard](https://github.com/mellinoe/veldrid/blob/33e5158046c708b50662fc8fda4a1f896d76a886/src/Veldrid/Vk/VkGraphicsDevice.cs#L668) [coded](https://github.com/mellinoe/veldrid/blob/33e5158046c708b50662fc8fda4a1f896d76a886/src/Veldrid/Vk/VkGraphicsDevice.cs#L742) in `VkGraphicsDevice`'s constructor. Is there a way I can enable more features?

TJYSunset:
Resolve #296 
Not sure I'm doing it right.

### DIFF
--- a/src/Veldrid/D3D11/D3D11GraphicsDevice.cs
+++ b/src/Veldrid/D3D11/D3D11GraphicsDevice.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Threading;
+using Feature = SharpDX.Direct3D11.Feature;
 
 namespace Veldrid.D3D11
 {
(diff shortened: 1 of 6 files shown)


mellinoe:
@TJYSunset Thanks for the contribution, this all looks good. I made some slight naming adjustments to the D3D11 implementation because SharpDX has been replaced by Vortice.

PR #302, opened 2020-09-02
D3D11 fix and leakage improvements

amerkoleci:
This fixes:
- ID3D11CommandList leakage with new method.
- ReportLiveObjects during D3D11 device dispose which was wrong.
- Staging buffers not being disposed.
- Calls non array ID3D11DeviceContext methods.


mellinoe:
Looks good. Thanks for the fixes!

PR #304, opened 2020-09-07
Fix ResolveTextureCore

trolleyman:
With the Vortice update looks like the src & dest parameters to the `ResolveSubresource` have been swapped around to more closely match the C api - now dest is first -- see https://github.com/amerkoleci/Vortice.Windows/blob/master/src/native/include/d3d12.h#L5530-L5535 vs https://github.com/sharpdx/SharpDX/blob/1f89381bc08f2d8cf8d294c6c5fce8dc85a849a1/Source/SharpDX.Direct3D11/DeviceContext.cs#L222-L224

This update switches the parameters around to fix this regression.

To check, try running NeoDemo & turning on MSAA in this commit vs master.


amerkoleci:
Nice catch, not sure why SharpDX didn't followed native API here and reversed the parameters, @mellinoe its good to merge for me.

trolleyman:
Yep it's a bit worrying, hopefully there are no other switcheroos like that in the codebase.

mellinoe:
LGTM, thanks for fixing this.

PR #307, opened 2020-09-15
Cap length of OpenGL debug names

trolleyman:
Ran into a small error where I set too long names for OpenGL - this limits the length of those names


mellinoe:
LGTM, thanks for the fix.

Issue #97, opened 2018-07-08
[Suggestion] Add more SDL2 functions


SamVanheer:
The Veldrid SDL2 library currently does not provide the functions to do things such as setting a window icon, which requires setting the image with [SDL_SetWindowIcon](https://wiki.libsdl.org/SDL_SetWindowIcon).

Functions like SDL_WaitEventTimeout and SDL_GetDisplayBounds are also missing.

Adding these would be nice, but would it be possible to make the functions used to load native functions public so these can be easily loaded without having to modify the SDL2 library?

AlzPatz:
How timely, I just started lookin into understanding the c++/c# Interop of this project with the aim of expanding a few things to include SDL input for joysticks and gamepads. 

I’m still trying to grok totally how this is working, but am I correct in thinking that NativeLibraryLoader removes the requirement for defining additional c callback signatures on the native side? Or is there still ‘wrapper’ work to do there, I.e if we need to expose a new function we need to update the native libraries? 

Thanks!

mellinoe:
> am I correct in thinking that NativeLibraryLoader removes the requirement for defining additional c callback signatures on the native side?

No, there's no getting around that. NativeLibraryLoader solves a different problem: loading a library that has different names on different platforms. SDL2 is a C library, so no work needs to be done within it to make the functions callable from C#.

AlzPatz:
Ah hah of course sdl2 is c not c++, I was being a bit dense but that now all makes sense! Thanks

SamVanheer:
@mellinoe Yes that would help. I'm currently using the SDL2-CS library since it already provides all of the functions in the native library, but it uses DllImport instead of NativeLibraryLoader to load the native library so it isn't quite as flexible.

RaZeR-RBI:
@mellinoe @SamVanheer I've created a [cross-platform SDL2 wrapper](https://github.com/project-grove/SDL2.NETCore) based on NativeLibraryLoader.
I haven't uploaded it to NuGet yet because I'm now focused on my [other project](https://github.com/project-grove/imagini) which uses that wrapper, but I'll happily do it if you need it now.

If it'll help you in any way, it'll be great :smiley:

PR #308, opened 2020-09-15
Add SDL_RaiseWindow

trolleyman:


zaafar:
FYI: https://github.com/mellinoe/veldrid/issues/97

trolleyman:
Right - so is the idea to only expose functions that are needed? I thought it might be cleaner for my code if all the SDL functions are in one place - i.e. in `Veldrid.SDL2`

zaafar:
yeah, looks like that's the idea.

mellinoe:
I don't have a problem with adding useful functions to Veldrid.SDL2, especially if they are windowing- or input-related.

PR #310, opened 2020-09-21
Add SDL_GetWindowDisplayIndex and SDL_GetCurrentDisplayMode

trolleyman:
Similar to #308, adds two extra functions that I plan to use to get the current display size.

`SDL_GetWindowDisplayIndex` is to get the display that that a window is currently on.

`SDL_GetCurrentDisplayMode` is to get the size of that display.

I don't want to use `SDL_GetDesktopDisplayMode` as according to the documentation the result of this method doesn't change when a window is fullscreen on the display with a different resolution, whereas I would want the changed resolution.

Also: unintentional, but it changes the line endings of the file to be uniformly LF. Can change back if wanted

### DIFF
--- a/src/Veldrid.SDL2/Sdl2.Window.cs
+++ b/src/Veldrid.SDL2/Sdl2.Window.cs
@@ -116,6 +116,16 @@ namespace Veldrid.Sdl2
         public static void SDL_SetWindowResizable(SDL_Window window, uint resizable) => s_setWindowResizable(window, resizable);
 
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+        private delegate int SDL_GetWindowDisplayIndex_t(SDL_Window window);
+        private static SDL_GetWindowDisplayIndex_t s_sdl_getWindowDisplayIndex = LoadFunction<SDL_GetWindowDisplayIndex_t>("SDL_GetWindowDisplayIndex");
+        public static int SDL_GetWindowDisplayIndex(SDL_Window window) => s_sdl_getWindowDisplayIndex(window);
+
+        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+        private delegate int SDL_GetCurrentDisplayMode_t(int displayIndex, SDL_DisplayMode* mode);
+        private static SDL_GetCurrentDisplayMode_t s_sdl_getCurrentDisplayMode = LoadFunction<SDL_GetCurrentDisplayMode_t>("SDL_GetCurrentDisplayMode");
+        public static int SDL_GetCurrentDisplayMode(int displayIndex, SDL_DisplayMode* mode) => s_sdl_getCurrentDisplayMode(displayIndex, mode);
+
+        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
         private delegate int SDL_GetDesktopDisplayMode_t(int displayIndex, SDL_DisplayMode* mode);
         private static SDL_GetDesktopDisplayMode_t s_sdl_getDesktopDisplayMode = LoadFunction<SDL_GetDesktopDisplayMode_t>("SDL_GetDesktopDisplayMode");
         public static int SDL_GetDesktopDisplayMode(int displayIndex, SDL_DisplayMode* mode) => s_sdl_getDesktopDisplayMode(displayIndex, mode);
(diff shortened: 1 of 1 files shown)


src/Veldrid.SDL2/Sdl2.Window.cs
```diff
@@ -213,14 +223,14 @@ public enum SDL_FullscreenMode : uint
...
-        public void* driverdata;
+    }
+
+    public unsafe struct SDL_DisplayMode
+    {
+        public uint format;
+        public int w;
+        public int h;
+        public int refresh_rate;
+        public void* driverdata;
```
zaafar, line 234: FYI (rather than using void*): https://github.com/project-grove/SDL2.NETCore/blob/9293cb70bcea600aca329357632816eeb4b0b116/SDL2/SDL_video.cs#L139

src/Veldrid.SDL2/Sdl2.Window.cs
```diff
@@ -113,16 +113,26 @@ public static unsafe partial class Sdl2Native
...
+
+        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+        private delegate int SDL_GetCurrentDisplayMode_t(int displayIndex, SDL_DisplayMode* mode);
+        private static SDL_GetCurrentDisplayMode_t s_sdl_getCurrentDisplayMode = LoadFunction<SDL_GetCurrentDisplayMode_t>("SDL_GetCurrentDisplayMode");
+        public static int SDL_GetCurrentDisplayMode(int displayIndex, SDL_DisplayMode* mode) => s_sdl_getCurrentDisplayMode(displayIndex, mode);
+
+        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+        private delegate int SDL_GetDesktopDisplayMode_t(int displayIndex, SDL_DisplayMode* mode);
+        private static SDL_GetDesktopDisplayMode_t s_sdl_getDesktopDisplayMode = LoadFunction<SDL_GetDesktopDisplayMode_t>("SDL_GetDesktopDisplayMode");
+        public static int SDL_GetDesktopDisplayMode(int displayIndex, SDL_DisplayMode* mode) => s_sdl_getDesktopDisplayMode(displayIndex, mode);
```
zaafar, line 131: also, rather than using pointers maybe use ref
https://github.com/project-grove/SDL2.NETCore/blob/9293cb70bcea600aca329357632816eeb4b0b116/SDL2/SDL_video.cs#L239

zaafar:
While we are here, I am using `SDL_GetDisplayBounds` to figure out the X, Y of a given display. I wonder if you would be okay if I add that too? @mellinoe

src/Veldrid.SDL2/Sdl2.Window.cs
trolleyman: I'm following the existing convention in the file, so to use ref instead of pointers I should probably change the other functions as well, right?

Not sure of what the C# convention is for pointers - `out`, `ref`, `*`, or `IntPtr`.

trolleyman:
Sure, I've added that to the MR. I've added it blind so we'll see if it actually compiles 😅

trolleyman:
...and this is why you don't blindly push things kids. I've got it working now.

src/Veldrid.SDL2/Sdl2.Window.cs
zaafar: oh, sorry I didn't see that!! yeah, since you are following existing convention, ignore my comment! :)

zaafar:
thx!

trolleyman:
@mellinoe Is this ok to merge?

mellinoe:
LGTM, thanks!

src/Veldrid.SDL2/Sdl2.Window.cs
```diff
@@ -213,14 +228,22 @@ public enum SDL_FullscreenMode : uint
...
+    {
+        public uint format;
+        public int w;
+        public int h;
+        public int refresh_rate;
+        public void* driverdata;
     }
 
-    public unsafe struct SDL_DisplayMode
+    public struct SDL_Rect
```
zaafar, line 242: @trolleyman  @mellinoe  I was going to use this new API but then I noticed that rather than using (https://github.com/mellinoe/veldrid/blob/bab9e25d808352fdcc36852ee7c7a1e059c3fe70/src/Veldrid.SDL2/Rectangle.cs) we have created a new struct (i.e. SDL_Rect). If there any reason why we are not using Rectangle.cs struct? If there is no reason would it be okay if I create a PR to use Rectangle rather than SDL_Rect and remove SDL_Rect ( if no one else is using it ).


P.S. Sorry for late comment (i.e. after PR is merged)

mellinoe: Good catch, I fixed that and it uses Rectangle now.

Issue #84, opened 2018-06-01
[Enhancement] Span<T> and Memory<T> usage


thargy:
With, [.Net Core 2.1 finally going live (and being LTS)](https://blogs.msdn.microsoft.com/dotnet/2018/05/30/announcing-net-core-2-1/), now seems the most appropriate time to consider some refactoring to liter the API with `Span<T>` and `Memory<T>`, considering this kind of project is really what they are made for and should deliver serious performance improvements.

mellinoe:
I'd like to add some extra overloads to support Span, but it's probably a good idea to wait for the release kinks to be worked out. I've seen problems on Xamarin in the past (when using ImageSharp), and @SomeAnon42 mentioned that there still seem to be a couple of minor issues being worked out:

https://github.com/xamarin/xamarin-android/issues/1679
https://github.com/dotnet/corefx/issues/29654

> this kind of project is really what they are made for and should deliver serious performance improvements

Realistically, they shouldn't help our performance too much, and I would consider adding them primarily for convenience. `Span<T>` is good for when you want to seek over a large-ish buffer and operate on different subsections without allocating. Veldrid doesn't really do that kind of thing, so it's not clear where we'd get any benefit. For the most part, Veldrid just takes a contiguous buffer you give it and passes it through to a native graphics API. We don't do much seeking and slicing. In any case, Span is certainly not faster than unsafe code (which is already prominently used because we are interfacing with native API's), but there may be places we can replace unsafe seeking with safe seeking via Spans and keep our performance.

The one place it could really help is the OBJ and MTL parsers (not part of core Veldrid), since those seek over a text file and parse through it. That's the perfect match for `Span<T>`, and it could help a ton in terms of perf and GC allocation. I had a version of ObjParser in the past which used `Span<T>`, but I ripped it out because Span wasn't stable and had issues on some platforms.

cwrea:
What @mellinoe said. Please, do get all of these new bits working across all .NET Standard platforms, first. There remain a number of issues on other platforms, all caused by recent changes for `Span<T>` to the `System.Memory` assembly:

- [aspnet/EntityFrameworkCore - Exception while loading assemblies - FileNotFoundException - Could not load assembly System.Memory #12081](https://github.com/aspnet/EntityFrameworkCore/issues/12081)

[...]

PR #311, opened 2020-09-26
Add span overload

feliwir:
Adresses #84


mellinoe:
Looks pretty good to me. It should be fine to use the “unmanaged” constraint here. The other overloads don’t use it because it didn’t exist at the time.

Can you add really basic tests that cover the new overloads? It should be sufficient to just copy one of the existing tests for the other overloads and use the Span version instead.

mellinoe (approved):

PR #313, opened 2020-10-12
Add Name property to GraphicsDevice

trolleyman:
This allows users of the library to access the debug name of the graphics device, so that it can be displayed to the user for debugging.

Tested every backend except for Metal, as I don't have a Mac. I see no reason why it shouldn't work though.


mellinoe:
I think this is a good addition, the only concern I have is that `Name` is a get/set property on most other Veldrid types that represents a debug name that the user can change, and this is a little different from that. I suggest changing the property to `AdapterName` to differentiate.

trolleyman:
Sure, that makes sense - cheers for the review!

mellinoe:
Thanks again. I've merged this and an additional change to include the name in NeoDemo's window title.
image

PR #317, opened 2020-11-06
Fixed redundant OpenGL vertex layout calls (#312)

### LINKED ISSUE #312: [OpenGL] Redundant calls to vertex layout setup code when pipeline and/or vertex buffer hasn't changed
When making multiple draw calls without changing pipeline or currently bound vertex buffers the command executor seems to be pushing out glBindBuffer and glVertexAttrib* calls before each draw call despite that there is no need to do so (if I understand correctly).

Here is how it looks in a trace:

<details>
<summary>Trace screenshot</summary>

Screenshot
</details>

[Relevant code is here](https://github.com/mellinoe/veldrid/blob/cdfaa674f069875e2c7758c01681ca20548f04b3/src/Veldrid/OpenGL/OpenGLCommandExecutor.cs#L255).

Seems like a simple solution to this would be making a boolean flag indicating whether we should update the vertex layouts and reset it at SetVertexBuffer(...) and SetPipeline(...), and do a flush in the PreDraw() method only if this flag is set.

I do have a patch ready and can send you a pull request if you don't mind. Of course, if I'm correctly understanding how this works and only if this change wouldn't break anything :wink:

<details>
<summary>Trace after the patch</summary>

Screenshot
(not sure why glGetError calls are not present in the first trace though)
</details>

Best regards!

RaZeR-RBI:
Fixes #312.

I don't have access to a development Windows machine at this moment (no bandwidth to set it up now) so I can't run the tests, but the NeoDemo seems to be working fine (Linux, OpenGL backend).

P.S.: Is there any plans for adding .gitattributes file to the repository? I've tried to commit the patch from a Linux machine, but it produced a full rewrite because the source file was misinterpreted as binary one because of the BOM mark at the start of the file.

Had to use the Github web interface to make this patch :neutral_face:

### DIFF
--- a/src/Veldrid/OpenGL/OpenGLCommandExecutor.cs
+++ b/src/Veldrid/OpenGL/OpenGLCommandExecutor.cs
@@ -35,6 +35,7 @@ namespace Veldrid.OpenGL
         private bool[] _newComputeResourceSets = Array.Empty<bool>();
 
         private bool _graphicsPipelineActive;
+        private bool _vertexLayoutFlushed;
 
         public OpenGLCommandExecutor(OpenGLGraphicsDevice gd, OpenGLPlatformInfo platformInfo)
         {
@@ -252,7 +253,11 @@ namespace Veldrid.OpenGL
             }
 
             FlushResourceSets(graphics: true);
-            FlushVertexLayouts();
+            if (!_vertexLayoutFlushed)
+            {
+                FlushVertexLayouts();
+                _vertexLayoutFlushed = true;
+            }
         }
 
         private void FlushResourceSets(bool graphics)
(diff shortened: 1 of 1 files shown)


RaZeR-RBI:
Got my hands on a machine with dev tools and an AMD card (R7 360), seems to be running fine (see screenshot).
I still can't run the tests though, many of them fail for some reason even on master for me - mostly D3D ones, but some OpenGL and Vulkan ones also. Some of them even fail with an ERROR_SUCCESS code (is that even an error? Duh...).

<details>
<summary>NeoDemo screenshot</summary>

veldrid-patch
</details>

mellinoe:
Some of the test failures are actually legitimate problems with the switch over to Vortice. I'm working on a fix for that and then will merge this in once I get a clean test run (or can confirm that this has no effect on any remaining test failures). Thanks for the improvements!

mellinoe:
@RaZeR-RBI I fixed the test failures, and merged this in. Thanks again!

Issue #324, opened 2020-12-10
WaitForFences() throws when timeout is infinite


bensmooth:
Calling [GraphicsDevice.WaitForFences()](https://github.com/mellinoe/veldrid/blob/46252eb6dfaba73b368be28366c4a8bfd96cffb9/src/Veldrid/D3D11/D3D11GraphicsDevice.cs#L529) with nanosecondTimeout=ulong.MaxValue causes the WaitHandle to throw because we are blindly converting ulong.MaxValue to ms by dividing by 1 million and then trying to stuff the result into an int.  This causes the int to overflow and become negative, making WaitHandle throw:
```
System.ArgumentOutOfRangeException: Number must be either non-negative and less than or equal to Int32.MaxValue or -1.
Parameter name: millisecondsTimeout
   at System.Threading.WaitHandle.WaitAll(WaitHandle[] waitHandles, Int32 millisecondsTimeout, Boolean exitContext)
   at Veldrid.D3D11.D3D11GraphicsDevice.WaitForFences(Fence[] fences, Boolean waitAll, UInt64 nanosecondTimeout)
   at Veldrid.GraphicsDevice.WaitForFences(Fence[] fences, Boolean waitAll)
```

From looking at the code, I believe this happens in all of the backends except for Vulkan.

mellinoe:
This should be fixed now; thanks for the PR.

PR #325, opened 2020-12-10
Fixed infinite WaitForFences throwing.

bensmooth:
Fix for https://github.com/mellinoe/veldrid/issues/324.  

Could be nice to have a ConvertNanosecondsToWaitHandleMs() somewhere accessible to all of them, but I wasn't sure where that would go.

### DIFF
--- a/src/Veldrid/GraphicsDevice.cs
+++ b/src/Veldrid/GraphicsDevice.cs
@@ -183,7 +183,7 @@ namespace Veldrid
         /// <param name="fences">An array of <see cref="Fence"/> objects to wait on.</param>
         /// <param name="waitAll">If true, then this method blocks until all of the given Fences become signaled.
         /// If false, then this method only waits until one of the Fences become signaled.</param>
-        /// <param name="nanosecondTimeout">A value in nanoseconds, indicating the maximum time to wait on the Fence.</param>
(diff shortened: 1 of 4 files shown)


src/Veldrid/GraphicsDevice.cs
```diff
@@ -183,7 +183,7 @@ public bool WaitForFences(Fence[] fences, bool waitAll, TimeSpan timeout)
         /// <param name="fences">An array of <see cref="Fence"/> objects to wait on.</param>
         /// <param name="waitAll">If true, then this method blocks until all of the given Fences become signaled.
         /// If false, then this method only waits until one of the Fences become signaled.</param>
-        /// <param name="nanosecondTimeout">A value in nanoseconds, indicating the maximum time to wait on the Fence.</param>
+        /// <param name="nanosecondTimeout">A value in nanoseconds, indicating the maximum time to wait on the Fence.  Pass ulong.MaxValue to wait indefinitely.</param>
```
bensmooth, line 186: We already do exactly this in `GraphicsDevice.WaitForFences(Fence[] fences, bool waitAll)`

mellinoe:
Looks good. Thanks for the fix, I'm merging it in now.

PR #327, opened 2020-12-24
Greatly optimized ObjParser

TechnologicalPizza:


TechnologicalPizza:
I thought that the texture coordinate parse ("vt" branch in `ObjParser.Process`) could try to read piece by piece. 
So if only one piece exists we get `Vector2(piece0, 0)` and when two pieces exist we get `Vector2(piece0, piece1)`. 
I'm unsure what to do if there are more than two pieces. The simple solution would be to just skip the overflow.
Don't know how feasible or even meaningful it would be to change the texture coordinate type to `Vector3` just to support that.

src/Veldrid.Utilities/FastParse.cs
```diff
@@ -0,0 +1,171 @@
+﻿using System;
+
+namespace Veldrid.Utilities
+{
+    public static class FastParse
```
mellinoe, line 5: I'd like to see if we can remove this file altogether, but if it needs to stay, it should be `internal`.

src/Veldrid.Utilities/ObjParser.cs
```diff
@@ -12,58 +11,111 @@ namespace Veldrid.Utilities
...
         private readonly ParseContext _pc = new ParseContext();
+        private char[] _readBuffer;
 
         /// <summary>
         /// Parses an <see cref="ObjFile"/> from the given raw text lines.
         /// </summary>
         /// <param name="lines">The text lines of the OBJ file.</param>
         /// <returns>A new <see cref="ObjFile"/>.</returns>
-        public ObjFile Parse(string[] lines)
+        public ObjFile Parse(IEnumerable<string> lines)
```
mellinoe, line 27: This is a breaking change to the method signature, and needs to be reverted to maintain backwards compatibility.

src/Veldrid.Utilities/ObjParser.cs
```diff
@@ -265,82 +324,105 @@ public ObjFile FinalizeFile()
...
+                piece1 = pieces.Current;
+
+                if (!pieces.MoveNext())
+                    goto Fail;
+                piece2 = pieces.Current;
+
+                if (!exact || !pieces.MoveNext())
+                    return;
+
+                Fail:
```
mellinoe, line 373: Please don't use goto or labels in this code.

src/Veldrid.Utilities/ObjParser.cs
```diff
@@ -265,82 +324,105 @@ public ObjFile FinalizeFile()
...
             {
-                try
-                {
-                    float x = float.Parse(xStr, CultureInfo.InvariantCulture);
-                    float y = float.Parse(yStr, CultureInfo.InvariantCulture);
-                    float z = float.Parse(zStr, CultureInfo.InvariantCulture);
+                if (FastParse.TryParseDouble(xStr, out double x) &&
+                    FastParse.TryParseDouble(yStr, out double y) &&
+                    FastParse.TryParseDouble(zStr, out double z))
+                    return new Vector3((float)x, (float)y, (float)z);
```
mellinoe, line 332: Please use brackets even around a single-line block.

src/Veldrid.Utilities/ObjParser.cs
```diff
@@ -84,140 +136,147 @@ private class ParseContext
...
-                if (slashSplit.Length != 1 && slashSplit.Length != 2 && slashSplit.Length != 3)
-                {
-                    throw CreateExceptionForWrongFaceCount(slashSplit.Length);
-                }
+                if (faceComponents.IsEmpty)
+                    throw CreateExceptionForWrongFaceCount();
 
-                int pos = ParseInt(slashSplit[0], "the first face position index");
+                int firstSlash = faceComponents.IndexOf(_slashChar);
+                ReadOnlySpan<char> firstPart = faceComponents.Slice(firstSlash + 1);
```
mellinoe, line 260: IMO, the naming here is a little confusing, because "firstPart" actually refers to the second part of the line in my thinking. Then, "firstSlice" is defined below, but that one makes sense for its name. This is effectively "afterFirstSlash" -- it's the Span for everything after the first slash.

src/Veldrid.Utilities/ObjParser.cs
```diff
@@ -372,33 +454,50 @@ public class ObjFile
...
         }
 
         /// <summary>
         /// Gets a <see cref="ConstructedMeshInfo"/> for the given OBJ <see cref="MeshGroup"/>.
         /// </summary>
         /// <param name="group">The OBJ <see cref="MeshGroup"/> to construct.</param>
+        /// <param name="reduce">Whether to simplify the mesh by sharing identical vertices.</param>
         /// <returns>A new <see cref="ConstructedMeshInfo"/>.</returns>
-        public ConstructedMeshInfo GetMesh(MeshGroup group)
+        public ConstructedMeshInfo GetMesh(MeshGroup group, bool reduce = true)
```
mellinoe, line 470: I tend to avoid default parameters in public API's. Can you change this to a required parameter and have the existing overload delegate to it w/ `true` ?

src/Veldrid.Utilities/ObjParser.cs
```diff
@@ -499,22 +597,53 @@ public MeshGroup(string name, string material, Face[] faces)
         /// <summary>
         /// An OBJ file construct describing the indices of vertex components.
         /// </summary>
-        public struct FaceVertex
+        public readonly struct FaceVertex : IEquatable<FaceVertex>
```
mellinoe, line 600: This is a breaking change and needs to be reverted.

src/Veldrid.Utilities/ObjParser.cs
```diff
@@ -525,7 +654,7 @@ public override string ToString()
         /// <summary>
         /// An OBJ file construct describing an individual mesh face.
         /// </summary>
-        public struct Face
+        public readonly struct Face
```
mellinoe, line 657: This is also a breaking change.

mellinoe (commented):
I'm supportive of making ObjParser faster, but I think this may go too far in micro-optimizing the code, which is making it more complicated, harder to review, and more difficult to maintain in the future. One thing to note is that I've just added a .NET Standard 2.1 target to this library, so we can use some of the built-in Span methods that were added, including `double.Parse(ReadOnlySpan<char>)`. I'd like to see if we can remove all of the FastParse stuff to start with. I'm skeptical of the ReadOnlySpanSplitter stuff as well, simply because it's a huge amount of code and very hard to trace through. Given that it's used in one place only, I suspect it would be shorter and easier to follow if the logic was just done within the Parse method.

Another problem right now is that there are no tests, functional or performance-wise, so it is hard to tell:
1. If this new logic is correct, including in error cases (malformed OBJ files, etc.)
and
2. If this is actually faster than before, and by how much. This is pretty important because there's some major readability/maintainability tradeoffs here, and to make a decision on those I need to know how much faster this is, objectively.

To alleviate this, we should add a couple of test projects. Unit tests can just use what Veldrid.Tests is using (Xunit), and the performance tests should use BenchmarkDotNet.

Also note that we shouldn't make any breaking changes in this PR. I've left a few comments, but we need to be careful about changing the public-facing API on anything, including adding "readonly" to a struct definition.

src/Veldrid.Utilities/ObjParser.cs
TechnologicalPizza: Can easily be reverted.

src/Veldrid.Utilities/ObjParser.cs
TechnologicalPizza: The `readonly` part can easily be reverted. The `IEquatable<FaceVertex>` removes millions of allocations as the mesh simplifier`s `Dictionary` cast every `FaceVertex` to `object` to check for equality.

src/Veldrid.Utilities/ObjParser.cs
TechnologicalPizza: I'll try to clarify

src/Veldrid.Utilities/ObjParser.cs
TechnologicalPizza: Easy fix

src/Veldrid.Utilities/FastParse.cs
TechnologicalPizza: Can be removed with the netstandard2.1 target

TechnologicalPizza:
Agree on the comments. I tried to replicate the behavior to be identical to the old parser (excluding the texture coordinate change).

Measured the performance using Performance Profiler in Visual Studio. Was around 2x of base and allocations were reduced to almost nothing. The `FastParse` class is a major speed contributor as it's faster than `System` methods, but the code bloat is overkill if we have access to the `Span` overloads from `System`.

src/Veldrid.Utilities/ObjParser.cs
mellinoe: That's fine, adding `readonly` is the only breaking change here, `IEquatable` can stay.

src/Veldrid.Utilities/ObjParser.cs
TechnologicalPizza: Does this rule apply to the whole class or just the `ExpectPieces` methods?

TechnologicalPizza:
I'm quite satisfied. Can't think of anything to improve on this.

mellinoe:
This is still missing some functional test cases and performance tests. I'm not expecting anything crazy there, just some basic validation and a way to prove that the new code is faster than before. The benchmarks could be as simple as timing how long it takes to load the same OBJ files from the functional tests.

TechnologicalPizza:
Ok so about tests. Should I copy the un-optimized ObjParser and then compare it to the one in the current branch? 
And presuming that NeoDemo is the "base" test as it's already in the repo; is that single model file enough, or should I add something more? I've looked at [this page](https://people.sc.fsu.edu/~jburkardt/data/obj/obj.html) (licensed under GNU LGPL). Don't know if we should include them in the repo, maybe downloading them into a cache on demand would be better? 

And about the performance tests. Should I include them in a commit, new branch, or just post some tables/stats in this PR?

mellinoe:
If you include the benchmarks as a separate commit then we can easily cherry-pick it into a branch with the older version and compare results. The change here doesn't impact the public API so we should be able to run any test against it as-is.

Starting with the obj files that are already in the repo (NeoDemo + Veldrid.VirtualReality.Sample) is fine with me. We can also add hand-crafted OBJ files if we want -- it doesn't really matter what the model actually looks like for the sake of a test case.

src/Veldrid.Utilities/ObjParser.cs
```diff
@@ -717,22 +750,22 @@ public struct Face
             /// <summary>
             /// The first vertex.
             /// </summary>
-            public readonly FaceVertex Vertex0;
+            public FaceVertex Vertex0;
```
TechnologicalPizza, line 753: Hopefully changing from readonly to normal field doesn't break compatibility. Otherwise the finalization code would need to be slightly more verbose.

TechnologicalPizza:
Still working on the tests though (didn't have much free time last couple days). 
And so I don't write weird asset loading into the test; should I add the assets as build-time outputs or get them with something along the lines of 
```cs
string loc = Assembly.GetExecutingAssembly().Location;
string src = Path.GetFullPath(Path.Combine(loc, "..", "..", "..", "..", "..", "src"));
string neodemo = Path.Combine(src, "NeoDemo");
``` though this may not be portable outside Windows

**Comment by TechnologicalPizza** [comment]
Quite promising benchmark results for SponzaAtrium.
<table>
<thead><tr><th>           Method</th><th>Mean</th><th>Error</th><th>StdDev</th><th>Gen 0</th><th>Gen 1</th><th>Gen 2</th><th>Allocated</th>
</tr>
</thead><tbody><tr><td>OriginalStream</td><td>1,119.0 ms</td><td>203.28 ms</td><td>11.14 ms</td><td>95000.0000</td><td>3000.0000</td><td>-</td><td>397.5 MB</td>
</tr><tr><td>OriginalLines</td><td>1,003.5 ms</td><td>177.38 ms</td><td>9.72 ms</td><td>73000.0000</td><td>2000.0000</td><td>-</td><td>262.11 MB</td>
</tr><tr><td>ImprovedStream</td><td>319.7 ms</td><td>25.28 ms</td><td>1.39 ms</td><td>1000.0000</td><td>-</td><td>-</td><td>29.49 MB</td>
</tr><tr><td>ImprovedLines</td><td>303.6 ms</td><td>32.37 ms</td><td>1.77 ms</td><td>500.0000</td><td>-</td><td>-</td><td>29.48 MB</td>
</tr><tr><td>ImprovedFastParseStream</td><td>182.1 ms</td><td>26.13 ms</td><td>1.43 ms</td><td>666.6667</td><td>333.3333</td><td>-</td><td>29.49 MB</td>
</tr><tr><td>ImprovedFastParseLines</td><td>166.3 ms</td><td>3.65 ms</td><td>0.20 ms</td><td>666.6667</td><td>333.3333</td><td>-</td><td>29.48 MB</td>
</tr></tbody></table>
Having the complicated FastParse helper is almost unnecessary now with the netstandard2.1 target.
I've actually made the ObjParser somewhat extensible in an upcoming commit which allows me to use FastParse for my own project to not bloat Veldrid.Utilities :rocket:    
  

<details>
<summary>Platform info</summary>
<pre><code>BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
Intel Core i7-4720HQ CPU 2.60GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=5.0.102
  [Host] : .NET Core 5.0.2 (CoreCLR 5.0.220.61120, CoreFX 5.0.220.61120), X64 RyuJIT
</code></pre>
<pre><code>Job=ShortRun  EvaluateOverhead=True  Toolchain=InProcessEmitToolchain  
IterationCount=3  LaunchCount=1  WarmupCount=3  
</code></pre>
</details>

**Comment by TechnologicalPizza** [comment]
Making most of the ParseContext methods virtual had no measurable performance impact and it allows for greatly customizable parsing (and allows me to remove the ObjParser copy from my project because I could just customize this new one :smile:)

File: src/Veldrid.Utilities/ObjParser.cs (hunk 9 above)
Comment (mellinoe) [reply src/Veldrid.Utilities/ObjParser.cs:796]: It does break compatibility, even if it is in a subtle way. How does the finalization code change?

**Review by mellinoe — CHANGES_REQUESTED** [review]
A lot more complexity and some breaking changes have crept into the latest version, which are making this into quite a large PR. The breaking changes need to be reverted (even though they are minor and unlikely to cause issues).

I would also like to revert the `virtual` changes. This might seem odd, but having the internals of the class exposed to the public make maintenance of the class a lot more difficult. My perspective is that if you are replacing almost all of the core parsing logic (in ParseContext), then you are better off writing a new OBJ parser that isn't tied to this interface. To reiterate, ObjParser wasn't originally designed to be very fast or flexible: it's just a really simple, dumb parser to get vertex data into example projects. If we can improve that core functionality without making it a ton more complicated, then I'm all for it. But I'd like to avoid going down a path that makes it significantly more complex, even if it would be more useful for other scenarios.

Those benchmark numbers look really good. Can you add that as a benchmark test project to be run next to Veldrid.Tests?

File: src/Veldrid.Tests/ObjParserTests.cs [hunk 10]
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        [Theory]
+        [MemberData(nameof(GetModelFiles))]
+        public void ParseStream(string file)
+        {
+            ObjParser parser = new ObjParser();
+            using (FileStream stream = File.OpenRead(file))
+            {
+                parser.Parse(stream);
```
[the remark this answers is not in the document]
TechnologicalPizza: Of course. This was more of a "creating the initial testing and things". 
I don't know what would be the best way to check for correct values though. Should I take some data samples or make a one-to-one equality check for Sponza?

src/Veldrid.Utilities/ObjParser.cs
```diff
@@ -25,52 +37,122 @@ public class ObjParser
...
+                if (readIndex > 0)
+                    _pc.Process(_readBuffer.AsSpan(0, readIndex));
             }
 
             _pc.EndOfFileReached();
             return _pc.FinalizeFile();
         }
 
-        private class ParseContext
+        public class ParseContext
```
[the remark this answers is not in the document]
TechnologicalPizza: I kind of changed it so it fit my current project (having a basic parser that can easily be extended) 😅
Quite easy to revert and shouldn't affect much AFAIK

TechnologicalPizza:
I'll try to improve tests when I have more free time on my hands, and thanks for your patience!

PR #331, opened 2021-01-03
Fixes for OpenVR

hyblocker:
Updated OpenVR API, and included the 32-bit version of OpenVR API
Fixed a bug where OpenVR API was not being loaded into memory, although a reimplementation might be necessary for it to work on Linux
Made `Veldrid.VirtualReality.Samples` more user friendly towards users who do not own a VR headset.


src/Veldrid.VirtualReality/Veldrid.VirtualReality.csproj
```diff
@@ -31,6 +31,18 @@
...
   </ItemGroup>
 
+
+  <ItemGroup>
+    <Folder Include="native\win-x86\" />
+  </ItemGroup>
+
+
+  <ItemGroup>
+    <None Update="native\win-x86\openvr_api.dll">
```
mellinoe, line 41: The x64 binary is included and packaged a few lines above, I think the x86 DLL should follow the same pattern. Currently, it won't be included in the NuGet package.

src/Veldrid.VirtualReality/OpenVR/OpenVRContext.cs
```diff
@@ -31,6 +31,24 @@ internal class OpenVRContext : VRContext
...
 
+        // The relative directories for the native binaries
+        private const string LibDir64 = "native/win-x64";
+        private const string LibDir32 = "native/win-x86";
+
+        // DllImport to load OpenVR native binary, might need to reimplement using NativeLibraryLoader, but given that we're loading a VR API, and most VR users use Windows, it may not be worth it
+        // Then again, this API is OpenVR, which is the only VR API other than OpenXR available on Linux, so it may be worth it
+        // TODO: Reimplement using NativeLibraryLoader
+        [System.Runtime.InteropServices.DllImport("kernel32.dll", CharSet = System.Runtime.InteropServices.CharSet.Auto, SetLastError = true)]
+        static extern bool SetDllDirectory(string lpPathName);
```
mellinoe, line 42: I don't think any of this should be necessary in theory. The generated openvr_api.cs file uses `DllImport` to invoke the initialization functions. Those DllImport's should resolve to the proper 32/64-bit binary automatically, assuming those binaries are properly packaged and referenced via NuGet. The 64-bit binary is already deployed to the expected `runtimes/win-x64/native` path, and the same should work for the 32-bit binary.

src/Veldrid.VirtualReality/OpenVR/OpenVRContext.cs
```diff
@@ -31,6 +31,24 @@ internal class OpenVRContext : VRContext
...
         internal GraphicsDevice GraphicsDevice => _gd;
 
+        // The relative directories for the native binaries
+        private const string LibDir64 = "native/win-x64";
+        private const string LibDir32 = "native/win-x86";
+
+        // DllImport to load OpenVR native binary, might need to reimplement using NativeLibraryLoader, but given that we're loading a VR API, and most VR users use Windows, it may not be worth it
+        // Then again, this API is OpenVR, which is the only VR API other than OpenXR available on Linux, so it may be worth it
+        // TODO: Reimplement using NativeLibraryLoader
+        [System.Runtime.InteropServices.DllImport("kernel32.dll", CharSet = System.Runtime.InteropServices.CharSet.Auto, SetLastError = true)]
```
mellinoe, line 41: If this needs to stay, add a `using System.Runtime.InteropServices` to the top.

src/Veldrid.VirtualReality/OpenVR/OpenVRContext.cs
```diff
@@ -31,6 +31,24 @@ internal class OpenVRContext : VRContext
...
+        static extern bool SetDllDirectory(string lpPathName);
+
+        //To try loading the OpenVR native binary from the native directory
+        static OpenVRContext()
+        {
+            string libName = Environment.Is64BitProcess ? LibDir64 : LibDir32;
+            // A lot of System.IO calls are used to ensure that the correct path to the native binaries folder is evaluated, and it shouldn't be that big a performance hit anyway given that we are doing this during startup
+            SetDllDirectory(System.IO.Path.GetFullPath(System.IO.Path.Combine(System.IO.Path.GetDirectoryName(System.Reflection.Assembly.GetEntryAssembly().Location), libName)));
```
mellinoe, line 49: Ditto; if this needs to stay, add a `using System.IO` and `using System.Reflection`.

mellinoe (changes requested):

src/Veldrid.VirtualReality/Veldrid.VirtualReality.csproj
```diff
@@ -14,21 +14,20 @@
...
     <PackageReference Include="Veldrid" Version="4.5.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(UseStableVersions)' != 'true'" >
+  <ItemGroup Condition="'$(UseStableVersions)' != 'true'">
     <ProjectReference Include="..\Veldrid\Veldrid.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="native\win-x64\" />
```
mellinoe, line 26: I don't think this works as expected. I think that a `Folder` property by itself is only going to make it show up in VS / other IDEs, but I don't think any of the binaries in it will be copied to the output directory.

For that, you need the `Content Include=""` and `CopyToOutputDirectory` snippets that were removed below. I was expecting to see the same kind of logic that already existed for `win-x64\openvr_api.dll` to be duplicated for `win-x86\openvr_api.dll`.

src/Veldrid.VirtualReality/OpenVR/OpenVRContext.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System;
-using System.Numerics;
 using System.Text;
+using System.Numerics;
```
mellinoe, line 3: nit: This seems like an unnecessary change (and it's less alphabetical than before).

mellinoe (changes requested):

mellinoe:
I do have my Oculus headset set up again, so I can test this locally when it's in a good state.

hyblocker:
Applied requested changes

PR #332, opened 2021-01-04
Update to latest Vortice.Windows 

amerkoleci:
Update to latest Vortice.Windows and map directly use Veldrid.Viewport in D3D11.


mellinoe:
Thanks for the update, LGTM.

PR #333, opened 2021-01-04
Resolved Isssue with using wrong variable for glBindFrameBuffer

voldien:
Use the created framebuffer object instead of a texture object when binding an OpenGL framebuffer.


mellinoe:
LGTM, thanks for the fix

PR #334, opened 2021-01-04
Improved SDL2 Native Pointer Extraction

voldien:
Added support for window native pointer and renamed some of the window info variables to reflect the name used in the SDL2 C library. The name was changed because the Sdl2Window suggests it is an SDLWindow pointer instead of the native window pointer.

### DIFF
--- a/src/Veldrid.SDL2/Sdl2.SysWMInfo.cs
+++ b/src/Veldrid.SDL2/Sdl2.SysWMInfo.cs
@@ -29,7 +29,7 @@ namespace Veldrid.Sdl2
         /// <summary>
         /// The Sdl2Window handle.
         /// </summary>
-        public IntPtr window;
+        public IntPtr Sdl2Window;
         /// <summary>
         /// The Sdl2Window device context.
         /// </summary>
@@ -43,7 +43,7 @@ namespace Veldrid.Sdl2
     public struct X11WindowInfo
     {
         public IntPtr display;
-        public IntPtr window;
+        public IntPtr Sdl2Window;
     }
 
     public struct WaylandWindowInfo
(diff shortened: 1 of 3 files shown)


src/Veldrid.SDL2/Sdl2.SysWMInfo.cs
```diff
@@ -43,7 +43,7 @@ public struct Win32WindowInfo
     public struct X11WindowInfo
     {
         public IntPtr display;
-        public IntPtr Sdl2Window;
+        public IntPtr window;
```
mellinoe, line 46: For better or worse, these typos are part of the public API, and can't be changed at this point.

mellinoe (changes requested):
Thanks for the submission! The improvements to `GetUnderlyingWindowHandle` look good to me, but the breaking changes to the public API need to be reverted for this to be mergeable. The name of those fields was obviously a mistake initially, but backwards compatibility is extremely important for the library.

src/Veldrid.SDL2/Sdl2.SysWMInfo.cs
voldien: Okay, would you like me to revert the variable names?

src/Veldrid.SDL2/Sdl2.SysWMInfo.cs
```diff
@@ -58,7 +58,13 @@ public struct CocoaWindowInfo
         /// <summary>
         /// The NSWindow* Cocoa window.
         /// </summary>
-        public IntPtr Window;
+        public IntPtr window;
```
mellinoe, line 61: This still slipped through in the second commit. I'll merge manually after fixing this.

mellinoe:
Merged manually, thanks for the improvements.

PR #335, opened 2021-01-05
Adding RenderDoc 1.4

voldien:
Added RenderDoc 1.4 API, but the application still uses version 1.2.


src/Veldrid.RenderDoc/RenderDocNative.cs
```diff
@@ -445,12 +447,18 @@ internal enum RENDERDOC_OverlayBits : uint
     // This will return 1 if the capture succeeded, and 0 if there was an error capturing.
     internal unsafe delegate uint pRENDERDOC_EndFrameCapture(void* device, void* wndHandle);
 
+    // Ends capturing immediately and discard any data stored without saving to disk.
+    //
+    // This will return 1 if the capture was discarded, and 0 if there was an error or no capture
+    // was in progress
+    internal unsafe delegate uint pRENDERDOC_DiscardFrameCapture(void* device, void* wndHandle);
```
mellinoe, line 454: This looks pretty harmless to expose publically, but it's not high priority.

mellinoe (approved):

PR #336, opened 2021-01-05
Allow non square compressed texture update support.

voldien:
Allow for none square compressed texture update to be supported. The OpenSage is crashing when updating non-square compressed textures.


mellinoe:
This looks like a good fix to me; thanks. Do you have a simple repro that we can also add as a test case to TextureTests.cs, based on what is crashing in OpenSAGE?

voldien:
Great, I do not have any test code associated with it. But I can perhaps later add some more unit tests.

PR #338, opened 2021-01-06
Removed unused code that issues a GL_INVALID_OPERATION

voldien:
In order to use the SDL_GL_GetAttribute, the OpenGL context has to have been created and make current. Which at this point has not been. But the code is never used so it should be removed.

### DIFF
--- a/src/Veldrid.StartupUtilities/VeldridStartup.cs
+++ b/src/Veldrid.StartupUtilities/VeldridStartup.cs
@@ -269,12 +269,7 @@ namespace Veldrid.StartupUtilities
                 }
             }
 
-            int actualDepthSize;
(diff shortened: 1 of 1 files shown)


voldien:
I believe it is related to https://github.com/OpenSAGE/OpenSAGE/issues/388

mellinoe:
SDL_GL_CreateContext is called on [line 260](https://github.com/mellinoe/veldrid/pull/338/files#diff-d19641625c40a5d11a0c620a8b036a628a7cae6df5df847a6f3a28de725e4fb9R260), and that function should both create the context and make it current. So I'm not sure that the description here is accurate. However, the code is indeed unused, so there's no reason to keep it around. I'd like to understand whether this is actually fixing the underlying issue that OpenSAGE is facing, though.

voldien:
Now that you mention, you're right, it should have been made current, I'll do some investigation if this is a MESA issue or something else.

PR #340, opened 2021-01-08
Add Opengl shader program Validation

voldien:
There has been an issue with OpenSAGE, https://github.com/OpenSAGE/OpenSAGE/issues/389. Where the screen is black. This seems to have been the cause that the OpenGL shader program was in an invalid state and thus never executed probably. This PR adds OpenGL validation before executing the program after lining for Debug build only. This is so it becomes easier to debug.

Though, I don't know if should add some additional options or just disable exception when finding an invalid OpenGL shader program since I don't know if this will cause a lot of programs to crash using veldrid.

### DIFF
--- a/src/Veldrid.OpenGLBindings/OpenGLNative.cs
+++ b/src/Veldrid.OpenGLBindings/OpenGLNative.cs
@@ -593,6 +593,11 @@ namespace Veldrid.OpenGLBinding
         public static void glUseProgram(uint program) => p_glUseProgram(program);
 
         [UnmanagedFunctionPointer(CallConv)]
+        private delegate void glValidateProgram_t(uint program);
+        private static glValidateProgram_t p_glValidateProgram;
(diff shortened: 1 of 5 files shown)


mellinoe (changes requested):
This is an interesting addition, but I think that it should be hidden behind `GraphicsDeviceOptions.Debug`, since it has the possibility of introducing driver-specific false positives and other GL driver weirdness. I think it fits nicely when enabled at development time, though.

Right now, `GraphicsDeviceOptions.Debug` isn't passed all the way down into OpenGLCommandExecutor, but it shouldn't be that difficult to do that.

voldien:
Alright, It makes sense for it to be used in debug mode only, I have added [Conditional("Debug")] on the ValidateProgram method, so it should only be called during debug build, and will be omitted in the release build, 

I take that with using the GraphicsDeviceOptions.Debug, is to make the validation optional during runtime in debug build.

PR #344, opened 2021-01-17
Add indirect draw indexed entry for OpenGL

voldien:
Making sure that the indirect draw indexed command is recorded for OpenGL.


mellinoe:
LGTM. Veldrid.Tests needs some test cases that use indirect drawing, apparently.

PR #350, opened 2021-01-22
Allow automatic disposals of objects

BSG-75:
Currently, `IDisposable` objects added to [GraphicsDevice.DisposeWhenIdle](https://github.com/mellinoe/veldrid/blob/c44982e59b0a77a3e01ece926e9993dc72986b0d/src/Veldrid/GraphicsDevice.cs#L721) will not be disposed until an explicit call to `GraphicsDevice.WaitForIdle` has been made.
This may cause high memory usage when it's not desired to call `WaitForIdle` too many times (since it will block), or in other cases, it "leaks memory" if people forget to call `WaitForIdle`.

This pull request doesn't change the behavior of `DisposeWhenIdle` (objects are not disposed, so guarenteed to be valid, until a `WaitForIdle()`), since there might be people that relies upon that behavior.
Instead, this pull request will add some new methods, in particular `GraphicsDevice.AutoDisposeWhenIdle`.
This new method behaves simiarly to `DisposeWhenIdle`, excepts that it might "decide" by itself to block the device until idle, and then automatically release resources.

We also had some discussion about it in the Veldrid Discord channel: https://discord.com/channels/757148685321895936/757148685321895939/801575253373222912

### DIFF
--- a/src/Veldrid/GraphicsDevice.cs
+++ b/src/Veldrid/GraphicsDevice.cs
@@ -13,6 +13,9 @@ namespace Veldrid
     {
         private readonly object _deferredDisposalLock = new object();
         private readonly List<IDisposable> _disposables = new List<IDisposable>();
+        private readonly List<IDisposable> _autoDisposables = new List<IDisposable>();
+        private long _autoDisposeThreshold;
+        private long _autoDisposeCounter;
         private Sampler _aniso4xSampler;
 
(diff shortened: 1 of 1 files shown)


TechnologicalPizza:
Just noticed that `DisposeWhenIdle` maybe should get a null check.

PR #352, opened 2021-01-29
Make SDL window title support Unicode characters

limocute:
Make SDL window title support Unicode characters


SimantoR:
Has anyone tested this?

mellinoe:
@limocute Thanks for the contribution!

The implementation here allocated a lot of memory that it didn't free (`Marshal.AllocHGlobal`) and just extra managed byte arrays. I reworked the change so that the C# string -> UTF8 marshalling happens inside the body of the method and uses stack allocation. Additionally, I've added the same logic to SDL_SetWindowTitle so that you can use a UTF8 string when adjusting the window title at runtime as well. I merged this change directly to master. Please give it a try and let me know if it works for your use case.

PR #353, opened 2021-01-30
[Vulkan] Fix IEnumerator allocations in CommandList.PreDispatchCommand.

Doom2fan:


mellinoe:
LGTM, thanks for the fix.

PR #355, opened 2021-02-01
Fixed D3D11 3D textures only able to write to depth=0

### LINKED ISSUE #351: Compute shaders only write to the first layer in a 3D texture in Direct3D11.
It seems that compute shaders with the D3D11 backend are unable to write to 3D textures past z=0.  The same code running with the Vulkan backend is able to write to the entire depth of the texture.  

With RenderDoc, I have verified that the output texture only gets populated for z=0 when using the D3D11 backend.  All of the other layers are filled with (0, 0, 0, 0).  Am I missing something, or is this a deficiency in the way Veldrid is interfacing with D3D11?  

Here is a ~200 line example project for reproduction:
https://github.com/bensmooth/Veldrid3dComputeShader
I make a 32x32x32 3D texture and run a dumb compute shader on it that just fills the texture with a value from a uniform.  Then, I copy each layer of the texture back and make sure it actually got filled properly.  With the D3D backend, only the first layer is properly filled.

Running runD3D.bat produces results I would call incorrect.  Running runVulkan.bat produces the results I would expect.  

Anyone have any hints as to where I can start digging around in Veldrid's D3D implementation to figure this out?

bensmooth:
Fixes #351.

### DIFF
--- a/src/Veldrid.Tests/ComputeTests.cs
+++ b/src/Veldrid.Tests/ComputeTests.cs
@@ -32,22 +32,24 @@ namespace Veldrid.Tests
         [Fact]
         public void ComputeShader3dTexture()
         {
-            // Just a dumb compute shader that fills a 3D texture with the same value from a uniform.
+            // Just a dumb compute shader that fills a 3D texture with the same value from a uniform multiplied by the depth.
             string shaderText = @"
 #version 450
 layout(set = 0, binding = 0, rgba32f) uniform image3D TextureToFill;
 layout(set = 0, binding = 1) uniform FillValueBuffer
 {
     float FillValue;
-    vec3 Padding_;
+    float Padding1;
+    float Padding2;
+    float Padding3;
 };
 layout(local_size_x = 16, local_size_y = 16, local_size_z = 1) in;
 void main()
 {
     ivec3 textureCoordinate = ivec3(gl_GlobalInvocationID.xyz);
-    vec4 dataToStore = vec4(FillValue);
+    float dataToStore = FillValue * (textureCoordinate.z + 1);
 
-    imageStore(TextureToFill, textureCoordinate, dataToStore);
+    imageStore(TextureToFill, textureCoordinate, vec4(dataToStore));
 }
 ";
 
@@ -102,9 +104,9 @@ void main()
             GD.WaitForIdle();
 
             // Read back from our texture and make sure it has been properly filled.
-            RgbaFloat expectedFillValue = new RgbaFloat(FillValue, FillValue, FillValue, FillValue);
             for (uint depth = 0; depth < computeTargetTexture.Depth; depth++)
             {
+                RgbaFloat expectedFillValue = new RgbaFloat(new System.Numerics.Vector4(FillValue * (depth + 1)));
                 int notFilledCount = CountTexelsNotFilledAtDepth(GD, computeTargetTexture, expectedFillValue, depth);
 
                 if (notFilledCount == 0)
--- a/src/Veldrid.Tests/ComputeTests.cs
+++ b/src/Veldrid.Tests/ComputeTests.cs
@@ -2,23 +2,194 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Text;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using Veldrid.SPIRV;
 using Xunit;
 
 namespace Veldrid.Tests
 {
     [StructLayout(LayoutKind.Sequential)]
-    public struct BasicComputeTestParams
+    struct FillValueStruct
     {
-        public uint Width;
-        public uint Height;
-        private uint _padding1;
-        private uint _padding2;
+        /// <summary>
+        /// The value we fill the 3d texture with.
+        /// </summary>
+        public float FillValue;
+        public float pad1, pad2, pad3;
+
+        public FillValueStruct(float fillValue)
+        {
+            FillValue = fillValue;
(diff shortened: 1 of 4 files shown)


src/Veldrid.Tests/ComputeTests.cs
```diff
@@ -2,23 +2,196 @@
...
+            FillValue = fillValue;
+            pad1 = pad2 = pad3 = 0;
+        }
     }
 
+
     public abstract class ComputeTests<T> : GraphicsDeviceTestBase<T> where T : GraphicsDeviceCreator
     {
+        [Fact]
+        public void ComputeShader3dTexture()
```
bensmooth, line 33: Added an integration test.  Both the Vulkan and Direct3D11 backends are able to pass the test after the change.  

The OGL backend fails.  I dug into this with renderdoc, and it looked like the 3D texture was getting populated correctly, and the staging textures were getting copied into correctly, but when we use MappableResource, all we get back is a bunch of zeros...

src/Veldrid/D3D11/D3D11TextureView.cs
```diff
@@ -71,8 +71,10 @@ public D3D11TextureView(D3D11GraphicsDevice gd, ref TextureViewDescription descr
                 {
                     uavDesc.ViewDimension = UnorderedAccessViewDimension.Texture3D;
                     uavDesc.Texture3D.MipSlice = (int)description.BaseMipLevel;
-                    uavDesc.Texture3D.FirstWSlice = (int)description.BaseArrayLayer;
-                    uavDesc.Texture3D.WSize = (int)description.ArrayLayers;
+
+                    // Map the entire range of the 3D texture.
+                    uavDesc.Texture3D.FirstWSlice = 0;
+                    uavDesc.Texture3D.WSize = (int)d3dTex.Depth;
```
bensmooth, line 77: The actual fix.  ArrayLayers is always going to be 1 for 3D textures, right?  I also think we want to map the whole 3D texture instead of starting the mapping at BaseArrayLayers.

src/Veldrid/Vk/VulkanUtil.cs
```diff
@@ -276,7 +276,9 @@ private static bool TryLoadVulkan()
             }
             else
             {
-                Debug.Fail("Invalid image layout transition.");
+                // TODO: Seems like we're missing a case for General to TransferSrcOptimal
+                // TODO: I think this is happening because the texture I have is being used in a compute shader.
+                //Debug.Fail($"Invalid image layout transition from '{oldLayout}' to '{newLayout}'.");
```
bensmooth, line 281: I had to comment this out in order to get the Vulkan test to pass.  Some vulkan wizard should probably figure out the correct flags for going from General to TransferSrcOptimal and add another case here.

src/Veldrid.Tests/ComputeTests.cs
```diff
@@ -2,23 +2,196 @@
...
+            // Read back from our texture and make sure it has been properly filled.
+            for (uint depth = 0; depth < computeTargetTexture.Depth; depth++)
+            {
+                RgbaFloat expectedFillValue = new RgbaFloat(new System.Numerics.Vector4(FillValue * (depth + 1)));
+                int notFilledCount = CountTexelsNotFilledAtDepth(GD, computeTargetTexture, expectedFillValue, depth);
+
+                if (notFilledCount == 0)
+                {
+                    // Expected behavior:
+                    Console.WriteLine($"All texels were properly set at depth {depth}");
```
mellinoe, line 115: Tests should be silent under normal operation.

src/Veldrid.Tests/ComputeTests.cs
```diff
@@ -2,23 +2,196 @@
...
+                }
+                else
+                {
+                    // Unexpected behavior:
+                    uint totalTexels = computeTargetTexture.Width * computeTargetTexture.Height;
+                    throw new Exception($"{notFilledCount} of {totalTexels} texels were not properly set at depth {depth}");
```
mellinoe, line 121: Use Xunit assertions so the tests fail in a standard way instead of throwing a generic Exception.

src/Veldrid.Tests/ComputeTests.cs
mellinoe: There must be a problem with an individual OpenGL codepath. The test succeeds on my setup.

src/Veldrid/Vk/VulkanUtil.cs
mellinoe: This should work as another branch of the logic:

```
            else if (oldLayout == VkImageLayout.General && newLayout == VkImageLayout.TransferSrcOptimal)
            {
                barrier.srcAccessMask = VkAccessFlags.ShaderWrite;
                barrier.dstAccessMask = VkAccessFlags.TransferRead;
                srcStageFlags = VkPipelineStageFlags.ComputeShader;
                dstStageFlags = VkPipelineStageFlags.Transfer;
            }
```

src/Veldrid/D3D11/D3D11TextureView.cs
mellinoe: The interface in Veldrid isn't super clear about how this is going to behave, and it's enough of an edge case that I haven't thought about it much. I think it's reasonable to treat the depth of a 3D texture as separate from the idea of an "array layer" although they are somewhat synonymous in other contexts. I think it's more natural to pass (0, 1) as (base array layer, layer count) than it would be to pass (0, depth). So this fix looks reasonable to me.

mellinoe:
I went ahead and merged this after applying some changes that I mentioned above. Thanks for the contribution!

PR #357, opened 2021-02-08
Add constant and slope-scaled depth bias options.

Doom2fan:
Adds depth bias support for all backends, except for Metal. D3D11, Vulkan and OpenGL appear to support it without extensions, and Metal should too, [according to this](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/1516269-setdepthbias).

I haven't added any tests, but it appears to work according to some testing with a hacked-up build of NeoDemo and with my own engine.

### DIFF
--- a/src/Veldrid/RasterizerStateDescription.cs
+++ b/src/Veldrid/RasterizerStateDescription.cs
@@ -27,6 +27,18 @@ namespace Veldrid
         /// Controls whether the scissor test is enabled.
         /// </summary>
         public bool ScissorTestEnabled;
+        /// <summary>
+        /// Controls whether depth bias is enabled.
+        /// </summary>
+        public bool DepthBiasEnabled;
+        /// <summary>
+        /// Controls how much constant depth bias to apply.
+        /// </summary>
+        public int DepthBiasConstant;
+        /// <summary>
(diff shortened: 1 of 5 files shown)


src/Veldrid/RasterizerStateDescription.cs
```diff
@@ -36,18 +48,27 @@ public struct RasterizerStateDescription : IEquatable<RasterizerStateDescription
...
+        /// <param name="depthBiasConstant">Controls how much constant depth bias to apply.</param>
+        /// <param name="depthBiasSlope">Controls how much slope-scaled depth bias to apply.</param>
         public RasterizerStateDescription(
             FaceCullMode cullMode,
             PolygonFillMode fillMode,
             FrontFace frontFace,
             bool depthClipEnabled,
-            bool scissorTestEnabled)
+            bool scissorTestEnabled,
+            bool depthBiasEnabled = false,
```
mellinoe, line 60: These parameters need to be added with a new constructor overload (and as non-optional params). The way these are added here is a binary breaking change, because the old overload effectively doesn't exist anymore.

mellinoe:
This needs a Metal implementation before it can be merged. Ideally, there would be a test too, but that might be tricky to implement given the effects of these options.

PR #359, opened 2021-02-20
supporting Cinematic4D exported obj with duplicate mat define and name with space

coolzoom:
This should be an issue but i rather just create an ugly pr so you know what i mean, although i tested it working but it may not the best way to achieve.

the problem is Cinematic4D would export the obj and mat file with space in the material name
```
mtl file:

newmtl mat_lionstatue copy
illum 1
map_Kd lionstatuecopy.png

obj file:

usemtl mat_lionstatue copy
f 36531/36531/36531 36532/36532/36532 36533/36533/36533
f 36533/36533/36533 36534/36534/36534 36535/36535/36535
f 36536/36536/36536 36537/36537/36537 36538/36538/36538
...
```

and inside the mtl file it may contains 

and it may have duplicated material name 
```
newmtl mat_mm_strmwnd_wall_01
illum 1
map_Kd mm_strmwnd_wall_01.png

...
newmtl mat_mm_strmwnd_wall_01
illum 1
map_Kd mm_strmwnd_wall_01.png

```


src/Veldrid.Utilities/MtlParser.cs
```diff
@@ -91,78 +91,99 @@ public void Process(string line)
                 switch (pieces[0].ToLowerInvariant().Trim())
                 {
                     case "newmtl":
-                        ExpectExactly(pieces, 1, "newmtl");
+                        ExpectAtLeast(pieces, 1, "newmtl");
                         FinalizeCurrentMaterial();
-                        _currentDefinition = new MaterialDefinition(pieces[1]);
+                        if (!DictMatDef.ContainsKey(pieces[1]))
```
mellinoe, line 96: There's a lot of diff here just because `_currentDefinition` might be null. I think it would be a lot simpler if you just did the duplication check over in `FinalizeCurrentMaterial`. IMO, you can also just get rid of the Dictionary altogether, and loop through the other materials and check their names in that method. There are going to be such a small number of materials in the majority of cases that it probably won't be any faster to use a Dictionary.

src/Veldrid.Utilities/MtlParser.cs
```diff
@@ -91,78 +91,99 @@ public void Process(string line)
...
+                        {
+                            DictMatDef.Add(pieces[1], pieces[1]);
+                            _currentDefinition = new MaterialDefinition(pieces[1]);
```
mellinoe, line 99: The name of the MaterialDefinition should be a concatenation of all pieces in the line, not just the first. Your example should give `"mat_lionstatue copy"`, and not just `"mat_lionstatue"`. This will matter if the MTL file, for whatever reason, included both `newmtl thing copy` and `newmtl thing copy 2`.

mellinoe:
I'm definitely supportive of fixing this. The OBJ/MTL specifications seem to be followed very loosely out in the wild, so I don't mind adding a bit of extra logic to handle corner cases like this.

PR #360, opened 2021-02-22
Add additional API surface for fine-grain control when veldrid is use…

seanspicer:
Add additional API surface for fine-grain control when veldrid is used with other Direct3D contexts.  This is important in order to be able to "hook" the device which Veldrid is using.  Without this change wildcard matches give inconsistent results.


mellinoe:

PR #363, opened 2021-02-28
Add swapchain readback support

SomeAnonDev:
I don't think it's currently possible to capture the contents of a Swapchain's Framebuffer. Using ``CopyTexture`` seems like the logical way to do it, and yet it fails with a crash. I added a [failing unit test](
630c357) that demonstrates exactly that.

This PR currently has fixes for D3D11 and Vulkan (see commit messages for more details). I don't have any idea how to implement this for the OpenGL backend, since it uses [placeholder textures](https://github.com/mellinoe/veldrid/blob/c0b2b821d59abfd1fde1f16918fd35e75f2ee5f0/src/Veldrid/OpenGL/OpenGLSwapchainFramebuffer.cs#L18) for swapchain framebuffer targets, and there's nothing you can do with those — certainly not pass them to ``CopyTexture``.
As for the Metal backend, there's just nothing I can do since I don't own a mac :man_shrugging:


mellinoe:
OpenGL needs to be implemented using [glReadPixels](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glReadPixels.xhtml) or a similar function. Metal should be feasible in a similar way to Vulkan. The CAMetalDrawable object has a `texture` property that could be copied from (in theory, haven't tested anything0.

SomeAnonDev:
image
Well, that wasn't too hard. I'm going to write a few more tests though.

SomeAnonDev:
Added OpenGL implementation and some more tests

mellinoe:
The implementation looks reasonable to me. I would like this to have a Metal implementation before merging, though. I'm able to test on my MacBook with the existing test if there's an implementation for it.

SomeAnonDev:
Okay, got it. There might be a problem with the Vulkan implementation though, specifically a missing layout transition. I've been using this code for a while in my project, and I've received at least one report that it crashes in ``TransitionImageLayout`` ("Invalid Image layout transition") whenever the game tries to capture the framebuffer. That person was using XWayland though, and I couldn't reproduce this on Windows.

src/Veldrid/OpenGL/OpenGLCommandExecutor.cs
```diff
@@ -1715,6 +1729,33 @@ public void CopyBuffer(DeviceBuffer source, uint sourceOffset, DeviceBuffer dest
             _stagingMemoryPool.Free(block);
         }
 
+        private void CopySwapchainFramebuffer(
+            OpenGLTexture dst,
+            uint srcX, uint srcY,
+            uint dstX, uint dstY, uint dstZ, uint dstMipLevel, uint dstLayer,
+            uint width, uint height)
+        {
+            glBindFramebuffer(FramebufferTarget.ReadFramebuffer, 0);
```
frenzibyte, line 1738: This should try using `SetSwapchainFramebuffer` first for platforms which have a swapchain framebuffer that's not zero (i.e. iOS).

PR #368, opened 2021-03-04
[Vulkan] Improved memory allocator

TechnologicalPizza:
The changes immensly reduce wasted space from allocating many small buffers by using different alignments based on buffer type.
Less memory is allocated per ChunkAllocator to reduce waste when there are a few allocations across different Vulkan memory types.
Freeing blocks now uses binary search which increases performance up to 5x when many free blocks are stored in one allocator.
Merging of blocks is now done when allocating and not when freeing, which massively improve exit times and reduces the amount of block merge attempts which reduces time spent on merging overall.

There is further tuning to be done around block size alignments and the usage per allocator.


TechnologicalPizza:
Ok, the only part left is to possibly improve chunk size constants.
I'm quite sure that [128 byte alignment](https://github.com/mellinoe/veldrid/pull/368/files#diff-f8b418617bdebacc0d5de60c2dc502d8562c6b98f33f891f188cf5c0a8adfcadR130) for chunks may not be optimal.

TechnologicalPizza:
Regarding [this constant](https://github.com/mellinoe/veldrid/blob/4dc7bffd6289056851f065ef5847df1da5f4fd87/src/Veldrid/Vk/VkGraphicsDevice.cs#L130), maybe it should be exposed through `VulkanDeviceOptions`?
1024 may still be low (4096 seems like a potential sweet spot, more testing needed). 
Having the ability to tune this constant would allow some nice memory savings when creating many small buffers (like I do in one of my projects). Observe that it would need to be round up to the nearest power of two because of the implementation.
This is not an alternative to pooling by the way; it just reduces overhead of small memory blocks.

PR #369, opened 2021-03-07
Added Span<T> overloads

TechnologicalPizza:
Just a quality of life improvement.


SimantoR:
Seems the .NET build fails due to myget feeds. Could you update it to fix those?

TechnologicalPizza:
I updated and now it fails for some other weird reason...

mellinoe:
Can you expand on when this would be useful? Wouldn't you be able to call the ReadOnlySpan in most cases since there is an implicit conversion?

TechnologicalPizza:
The implicit conversion doesn't kick in because ReadOnlySpan is generic and the type is not being inferred.
I find this annoying because the XML doc has to be duplicated in source but it's annoying otherwise.

mellinoe:
Got it. Yeah, after testing it, the generic nature of the method makes it so that it can't select the right overload. The only way to get it to work is to explicitly specify the type to `UpdateBuffer<T>` or to cast it manually. Neither is good. So this change looks correct to me.

PR #371, opened 2021-03-09
OperatingSystem.IsAndroid()

SupinePandora43:
Adds NET 5.0 target framework

uses `OperatingSystem.IsAndroid` for android checks

### DIFF
--- a/src/Veldrid/MTL/MTLGraphicsDevice.cs
+++ b/src/Veldrid/MTL/MTLGraphicsDevice.cs
@@ -8,6 +8,10 @@ using System.Threading;
 using NativeLibraryLoader;
 using Veldrid.MetalBindings;
 
+#if NET5_0_OR_GREATER
+using NativeLibrary = NativeLibraryLoader.NativeLibrary;
+#endif
+
 namespace Veldrid.MTL
 {
     internal unsafe class MTLGraphicsDevice : GraphicsDevice
--- a/src/Veldrid/MTL/MTLGraphicsDevice.cs
+++ b/src/Veldrid/MTL/MTLGraphicsDevice.cs
@@ -7,6 +7,9 @@ using System.Runtime.InteropServices;
 using System.Threading;
 using NativeLibraryLoader;
 using Veldrid.MetalBindings;
+#if NET5_0
+using NativeLibrary = NativeLibraryLoader.NativeLibrary;
+#endif
 
 namespace Veldrid.MTL
 {
--- a/src/Veldrid/MTL/MTLGraphicsDevice.cs
+++ b/src/Veldrid/MTL/MTLGraphicsDevice.cs
@@ -7,9 +7,6 @@ using System.Runtime.InteropServices;
 using System.Threading;
 using NativeLibraryLoader;
 using Veldrid.MetalBindings;
-#if NET5_0
-using NativeLibrary = NativeLibraryLoader.NativeLibrary;
-#endif
 
 namespace Veldrid.MTL
 {
--- a/src/Veldrid/MTL/MTLGraphicsDevice.cs
+++ b/src/Veldrid/MTL/MTLGraphicsDevice.cs
@@ -5,12 +5,8 @@ using System.IO;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Threading;
-using NativeLibraryLoader;
-using Veldrid.MetalBindings;
-
-#if NET5_0_OR_GREATER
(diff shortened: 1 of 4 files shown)


zaafar:
what's up with so many white space changes. :D

SupinePandora43:
> what's up with so many white space changes. :D

Vscode automatically normalized line endings using `.editorconfig`

SupinePandora43:
this is not needed if https://github.com/dotnet/runtime/issues/51052 will be fixed (when `RuntimeInformation.IsOSPlatform(OSPlatform.Linux)` will return `true`)

TechnologicalPizza (commented):
Seems like we may want to deprecate the old NativeLibraryLoader as it breaks horribly for publishes with embedded native dlls.

src/Veldrid/Vk/VkGraphicsDevice.cs
```diff
@@ -1,1519 +1,1523 @@
...
+            }
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                if (availableInstanceExtensions.Contains(CommonStrings.VK_KHR_WIN32_SURFACE_EXTENSION_NAME))
+                {
+                    _surfaceExtensions.Add(CommonStrings.VK_KHR_WIN32_SURFACE_EXTENSION_NAME);
+                }
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux)
```
TechnologicalPizza, line 471: We may want to change the use of `RuntimeInformation` throughout the whole library at once. The solution here is not future-proof (checking for NET5_0 will not work for NET6).
We may just want to create a NET6 branch of Veldrid 🤔

SupinePandora43: `OperatingSystem.IsAndroid()` was introduced in net5, and works as expected in net6.
In other words: net5 added, net6 implemented.

TechnologicalPizza: Well I meant the #if directive in particular. I'm not sure the `NET5_0` symbol will be available in NET6 (the `NET` symbol should exist though)

SupinePandora43: `NET5_0_OR_GREATER`

src/Veldrid/MTL/MTLGraphicsDevice.cs
```diff
@@ -8,6 +8,10 @@
 using NativeLibraryLoader;
 using Veldrid.MetalBindings;
 
+#if NET5_0_OR_GREATER
+using NativeLibrary = NativeLibraryLoader.NativeLibrary;
```
mellinoe, line 12: I don't think this needs to be behind an `#if`. It should be harmless to include everywhere.

src/Veldrid/MTL/MTLGraphicsDevice.cs
```diff
@@ -16,7 +20,7 @@ internal unsafe class MTLGraphicsDevice : GraphicsDevice
         private static readonly Dictionary<IntPtr, MTLGraphicsDevice> s_aotRegisteredBlocks
             = new Dictionary<IntPtr, MTLGraphicsDevice>();
 
-        private readonly MTLDevice _device;
+        private readonly MTLDevice _device;
```
mellinoe, line 23: There's a lot of unnecessary whitespace changes included in this diff. I can't tell what was actually changed. Can you remove these unexpected diffs from the PR?

src/Veldrid/MTL/MTLGraphicsDevice.cs
```diff
@@ -8,6 +8,10 @@
 using NativeLibraryLoader;
```
mellinoe, line 8: I think this may not be needed if the addition below is made everywhere (instead of behind the #if).

mellinoe (changes requested):
The functional changes look fine, please clean up some of the misc. things I commented on and we can merge.

src/Veldrid/MTL/MTLGraphicsDevice.cs
SupinePandora43: The previous attempt had much more space changes because editor used `.editorconfig`

SupinePandora43: @mellinoe i can't
image
^ this is what i get when trying to edit files on `master` branch.
those space changes are auto fixes by editor

mellinoe:
I merged this after rebasing it on master. Thanks for the contribution!

PR #373, opened 2021-04-02
Use Pipeline ResourceSetCount Instead of Array Length

generatives:
I found an error in VkCommandList caused by using two Compute Pipelines in the same CommandList. When dispatching the pipeline with fewer resources the loop inside VkCommandList.PreDispatchCommand would throw an exception because it would loop past the number of resource sets that had actually been bound. I have changed VkCommandList to use the number of resources in the current pipeline instead of the length of the array.


mellinoe:

PR #377, opened 2021-04-19
Update to latest Vortice.Windows

amerkoleci:
- There was bug with SharpGen during some AddRef and Release marshal of IUnknow.
- Update to latest Vortice.Windows and handle ReportLiveObjects in better way.


mellinoe:
LGTM, ty @amerkoleci

Issue #379, opened 2021-04-19
[Vulkan] Slow performance of GPU->CPU copy


baSSiLL:
My code takes results of rendering from a render target to a CPU memory using an intermediate staging texture. I noticed that the frame rate I can get with the Vulkan backend on Windows is substantially (like 10 times) lower than with the Direct3D backend. Narrowing down the problem revealed the bottleneck is a code that copies from the mapped memory of a staging texture to the CPU memory. Which is actually a simple call to **System.Buffer.MemoryCopy**.
After digging a bit I've discovered the **VK_MEMORY_PROPERTY_HOST_CACHED_BIT** flag. I just tried to use it in the **VkTexture** constructor where a memory for a staging buffer is allocated (instead of **VK_MEMORY_PROPERTY_HOST_COHERENT_BIT**). And viola - the performance of copying is on par with Direct3D, and I can get comparable frame rate with both backends and didn't see any drawbacks.
```c#
// line 187 of VkTexture.cs 
                _memoryBlock = _gd.MemoryManager.Allocate(
                    _gd.PhysicalDeviceMemProperties,
                    bufferMemReqs.memoryTypeBits,
                    VkMemoryPropertyFlags.HostVisible | VkMemoryPropertyFlags.HostCached,
                    true,
                    bufferMemReqs.size,
                    bufferMemReqs.alignment,
                    prefersDedicatedAllocation,
                    VkImage.Null,
                    _stagingBuffer);
```
It's the first ever time I deal with Vulkan thus I don't feel confident when it comes to working with it. Hence no pull request (yet).
Whether using "host cached" memory is a good idea for all staging resources? Probably, just for read only? Does it make sense to allow an explicit control over this flag for the Vulkan backend?

baSSiLL:
Nice presentation describing various memory types in Vulkan and what to use them for
https://www.khronos.org/assets/uploads/developers/library/2018-vulkan-devday/03-Memory.pdf

mellinoe:
I believe the code should just be updated to use a HOST_CACHED allocation if there exists a heap supporting it by the device, which is not guaranteed, and falling back to the current behavior otherwise.

> Whether using "host cached" memory is a good idea for all staging resources? Probably, just for read by CPU?

[...]

PR #380, opened 2021-04-22
[Vulkan] Use "host cached" memory for staging textures and buffers

baSSiLL:
This improves performance of GPU -> CPU transfers.
Related discussion can be found in #379

### DIFF
--- a/src/Veldrid/Vk/VkTexture.cs
+++ b/src/Veldrid/Vk/VkTexture.cs
@@ -185,10 +185,16 @@ namespace Veldrid.Vk
                     prefersDedicatedAllocation = false;
                 }
 
+                // Use "host cached" memory when available, for better performance of GPU -> CPU transfers
+                var propertyFlags = VkMemoryPropertyFlags.HostVisible | VkMemoryPropertyFlags.HostCoherent | VkMemoryPropertyFlags.HostCached;
+                if (TryFindMemoryType(_gd.PhysicalDeviceMemProperties, bufferMemReqs.memoryTypeBits, propertyFlags) == null)
+                {
+                    propertyFlags ^= VkMemoryPropertyFlags.HostCached;
+                }
                 _memoryBlock = _gd.MemoryManager.Allocate(
                     _gd.PhysicalDeviceMemProperties,
                     bufferMemReqs.memoryTypeBits,
-                    VkMemoryPropertyFlags.HostVisible | VkMemoryPropertyFlags.HostCoherent,
+                    propertyFlags,
                     true,
                     bufferMemReqs.size,
                     bufferMemReqs.alignment,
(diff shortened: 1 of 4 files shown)


src/Veldrid/Vk/VulkanUtil.cs
```diff
@@ -19,7 +19,7 @@ public static void CheckResult(VkResult result)
             }
         }
 
-        public static uint FindMemoryType(VkPhysicalDeviceMemoryProperties memProperties, uint typeFilter, VkMemoryPropertyFlags properties)
+        public static uint? TryFindMemoryType(VkPhysicalDeviceMemoryProperties memProperties, uint typeFilter, VkMemoryPropertyFlags properties)
```
zaafar, line 22: Should probably return a bool if you want to name the function tryXXX. (Similar to dictionary TryXXX functions). Also, use *out* parameter to return the actual result.

zaafar: I see your point, make sense.

Yeah, FindMemoryType sounds better than TryFindMemoryType but that’s personal preference I think. :)

baSSiLL:
@mellinoe please check if this pull request looks OK to you.

src/Veldrid/Vk/VulkanUtil.cs
mellinoe: I agree with the original suggestion here. `TryXYZ` is a well-understood pattern that this clashes with. It's also used outside of Dictionaries even in the BCL, see `int.TryParse` and friends.


To avoid changing all of the existing callers of `FindMemoryType`, you can just reformulate it in terms of the Try variant (abridged):


```C#
uint FindMemoryType(...) 
{
    if (!TryFindMemoryType(..., out memType)) throw VeldridException();
    else return memType;
}
```

src/Veldrid/Vk/VkTexture.cs
```diff
@@ -185,10 +185,16 @@ internal VkTexture(VkGraphicsDevice gd, ref TextureDescription description)
                     prefersDedicatedAllocation = false;
                 }
 
+                // Use "host cached" memory when available, for better performance of GPU -> CPU transfers
+                var propertyFlags = VkMemoryPropertyFlags.HostVisible | VkMemoryPropertyFlags.HostCoherent | VkMemoryPropertyFlags.HostCached;
+                if (TryFindMemoryType(_gd.PhysicalDeviceMemProperties, bufferMemReqs.memoryTypeBits, propertyFlags) == null)
```
mellinoe, line 190: The comparison to `null` here does feel awkward in the presence of a "Try" function. When reading snippets like this, I expect the "Try" function to return a boolean.

mellinoe (changes requested):

mellinoe (approved):
LGTM

baSSiLL:
Great @mellinoe
So why haven't you merged it then?

mellinoe:
@baSSiLL I was simply waiting for the test build to complete running.

baSSiLL:
OK, thanks.
Please forgive my persistence, didn't realize that you've approved just a few minutes ago :)

PR #384, opened 2021-05-07
[OpenGL] Make mapping buffer coherent

TechnologicalPizza:
Not marking Write-only buffers as Persistent/Coherent created massive overhead on AMD and Nvidia cards. 
Either mapping or unmapping (depending on vendor) caused massive performance degradation.

Before fix: mapping/unmapping a 64MB buffer blocked the OpenGL thread for ~50ms.
After fix: Specifying `BufferAccessMask.Coherent` when mapping results in performance comparable to Vulkan and DX11.

### DIFF
--- a/src/Veldrid/OpenGL/OpenGLFormats.cs
+++ b/src/Veldrid/OpenGL/OpenGLFormats.cs
@@ -576,7 +576,7 @@ namespace Veldrid.OpenGL
                 case MapMode.Read:
                     return BufferAccessMask.Read;
                 case MapMode.Write:
-                    return BufferAccessMask.Write | BufferAccessMask.InvalidateBuffer;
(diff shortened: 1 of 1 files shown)


TechnologicalPizza:
Seems like a AMD device of mine is throwing a InvalidOperation error when using the Coherent flag. Reading further into the spec it seems that Coherent/Persistent requires the memory to be explicitly initialized, yet that initialization method is only available for OpenGL 4.4 (or through ARB_buffer_storage). 
And as Veldrid does not specify whether automatic synchronization is required for mapped resources, Persistent can be used instead of Coherent (which could be a bit more performant).

TechnologicalPizza:
Seems like the best solution for now is to use Dynamic instead of Staging buffers, as Staging buffers have certain requirements on at least OpenGL (like being marked as Writeable or Readable upon creation, which makes mapping with different MapModes harder to implement).

PR #385, opened 2021-05-22
[D3D11] Fix independent blends not being enabled.

Doom2fan:
This PR fixes the "IndependentBlendEnable" boolean in the D3D11 blend state not being set to true.

### DIFF
--- a/src/Veldrid/D3D11/D3D11ResourceCache.cs
+++ b/src/Veldrid/D3D11/D3D11ResourceCache.cs
@@ -82,6 +82,7 @@ namespace Veldrid.D3D11
             }
 
             d3dBlendStateDesc.AlphaToCoverageEnable = description.AlphaToCoverageEnabled;
+            d3dBlendStateDesc.IndependentBlendEnable = true;
 
             return _device.CreateBlendState(d3dBlendStateDesc);
         }


jhm-ciberman:
Explanation: 
In my game I use two render targets. 
image

In Vulkan and OpenGL, this works fine:
image

But in DirectX11 the blending state for the second target is set incorrectly:
image

Also, the Independent blend is disabled in DirectX11:
image

And Veldrid is not setting this correctly in the BlendState as shown here: 
image

jhm-ciberman (commented):
It is possible to use the already existing for loop and check whether all `renderTargets[i].IsBlendEnabled` are equal and then set `independentBlendEnable` to that value? 

Something like:

```csharp
bool independentBlend = false;
bool previousIsBlendEnabled = attachamentStates.Length > 0 ? attachmentStates[0].IsBlendEnabled : false;
for (int i = 0; i < attachmentStates.Length; i++)
{
    BlendAttachmentDescription state = attachmentStates[i];
    independentBlend = independentBlend || (state.BlendEnabled != previousIsBlendEnabled);
    
    // ...
}

d3dBlendStateDesc.IndependentBlendEnable = independentBlend;
```

Correct me if I'm wrong. I'm not fully with DirectX api

mellinoe (approved):

PR #390, opened 2021-05-28
Exposed some backend information; useful for debugging

TechnologicalPizza:


MarkWilds:
Looks good and read only.

src/Veldrid/GraphicsApiVersion.cs
```diff
@@ -0,0 +1,67 @@
...
+        ///   </item>
+        ///   <item>
+        ///     <description>major_number.minor_number.release_number</description>
+        ///   </item>
+        /// </list>
+        /// </summary>
+        /// <param name="versionString">The OpenGL version string.</param>
+        /// <param name="version">The parsed <see cref="GraphicsApiVersion"/>.</param>
+        /// <returns>True whether the parse succeeded; otherwise false.</returns>
+        public static bool TryParseGLVersion(string versionString, out GraphicsApiVersion version)
```
mellinoe, line 42: I don't think this needs to be a public method.

src/Veldrid/MTL/MTLFeatureSupport.cs
```diff
@@ -1,17 +1,20 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using Veldrid.MetalBindings;
 
 namespace Veldrid.MTL
 {
-    internal class MTLFeatureSupport
+    public class MTLFeatureSupport : IReadOnlyCollection<MTLFeatureSet>
```
mellinoe, line 8: I'm not sure about directly exposing this as a public type. Some of the things here will feel a little strange to the average user: they only exist because they are interesting or useful to Veldrid's implementation. I'd actually be more comfortable directly exposing `_supportedFeatureSets` since it's directly from Metal.

src/Veldrid/OpenGL/OpenGLExtensions.cs
```diff
@@ -1,16 +1,18 @@
-﻿using System;
+﻿using System.Collections;
 using System.Collections.Generic;
 
 namespace Veldrid.OpenGL
 {
-    internal class OpenGLExtensions
+    public class OpenGLExtensions : IReadOnlyCollection<string>
```
mellinoe, line 6: Again, I'm not sure about exposing this whole type publically. The raw list of extensions should be available, but all of the other helper members on this type aren't a good idea to expose IMO.

src/Veldrid/BackendInfoOpenGL.cs
```diff
@@ -18,6 +19,51 @@ internal BackendInfoOpenGL(OpenGLGraphicsDevice gd)
...
+        /// Vendor-specific information may follow the version number.
+        /// Its format depends on the implementation, but a space always separates the version number and the vendor-specific information.
+        /// </para>
+        /// </remarks>
+        public string ShadingLanguageVersion => _gd.ShadingLanguageVersion;
+
+        /// <summary>
+        /// Gets a collection of available OpenGL extensions.
+        /// </summary>
+        public OpenGLExtensions Extensions => _gd.Extensions;
```
mellinoe, line 65: See other comment, but I think this should just be a `List<string>`, or a `ReadOnlyCollection<string>`, however similar things are being exposed. I don't think `OpenGLExtensions` itself needs to become public.

mellinoe (changes requested):
Good work overall, I'm definitely in support of this direction. I'm always concerned about exposing previously-internal types, and I think we can accomplish what we want here without making too much public.

src/Veldrid/GraphicsApiVersion.cs
TechnologicalPizza: Hmm I had a reason for making this public but I forgot it. Anyways, this parses based on OpenGL specification so it shouldn't ever change behavior-wise.

src/Veldrid/MTL/MTLFeatureSupport.cs
TechnologicalPizza: I actually like the IReadOnlyCollection idea you have with OpenGL extensions. The only useful thing `MTLFeatureSupport` provides though is the `MaxFeatureSet` property, which on second thought can simply be added to `BackendInfoMetal`

TechnologicalPizza:
Ok VkGraphicsDevice got full of whitespace changes so sorry about that.
Luckily there is a feature for mitigating that problem, so if you don't think it's necessary to fix the whitespaces you can use this:
image

mellinoe:
@TechnologicalPizza I'm looking at this again and it seems like a lot of unrelated changes got mixed into this branch which need to be pulled out.

TechnologicalPizza:
Oh yeah how strange. I'll try fixing it.

src/Veldrid/MTL/MTLGraphicsDevice.cs
```diff
@@ -52,15 +54,20 @@ internal unsafe class MTLGraphicsDevice : GraphicsDevice
...
+
             Features = new GraphicsDeviceFeatures(
                 computeShader: true,
                 geometryShader: false,
                 tessellationShaders: false,
                 multipleViewports: MetalFeatures.IsSupported(MTLFeatureSet.macOS_GPUFamily1_v3),
                 samplerLodBias: false,
-                drawBaseVertex: true,
-                drawBaseInstance: true,
+                drawBaseVertex: true,   // TODO?: MetalFeatures.IsDrawBaseVertexInstanceSupported()
```
mellinoe, line 69: I think these should be changed, based solely on this table: https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf

mellinoe: I made this change during the push to master.

mellinoe:
Looks good to me, merged it to master.

PR #393, opened 2021-06-17
Apply dotnet format changes

feliwir:
Run [dotnet-format](https://github.com/dotnet/format), which applies the rule from the `.editorconfig` to the entire solution. Mostly just small things


src/Veldrid.ImGui/ImGuiRenderer.cs
```diff
@@ -259,33 +259,33 @@ public void ClearCachedImageResources()
             switch (factory.BackendType)
             {
                 case GraphicsBackend.Direct3D11:
-                {
-                    if (stage == ShaderStages.Vertex && colorSpaceHandling == ColorSpaceHandling.Legacy) { name += "-legacy"; }
-                    string resourceName = name + ".hlsl.bytes";
-                    return GetEmbeddedResourceBytes(resourceName);
-                }
+                    {
```
mellinoe, line 262: Can you please modify the .editorconfig so that this indentation is the standard?

mellinoe (changes requested):

PR #394, opened 2021-06-17
Format ci

feliwir:
Run [dotnet-format](https://github.com/dotnet/format), with the `--check` argument which fails if it finds any files violating the style rules. See https://github.com/feliwir/veldrid/pull/1 for an example (only **Buffer.cs** is bad formatted). I was looking into more sophisticated actions e.g. https://github.com/marketplace/actions/dotnet-format , but these have some issues currently. 

Using those would allow only checking the diff and commenting on the PR. Question is if we should commit the changes immediatly.


feliwir:
Requires #393 to pass of course

PR #396, opened 2021-06-30
[Vulkan] Replaced ResourceRefCount with VkRefCountManager

TechnologicalPizza:
ResourceRefCount is a handy utility *class* used to track reference counts for Vulkan objects. It manages the reference count and calls an action upon reaching zero. While this design works well, it introduces big overhead due to CLR internals; reference types share generic instantiations and accessing arrays of reference types requires special checks due to covariance.

image
*Frame time currently: 29.5ms (34fps)
Highlighted methods are where most of the time is spent by the* **current** *reference-counting system, which include extremely common/hot paths like setting resource sets and buffers.*  
  
  
image
*Frame time with changes: 21.7ms (46fps)
Highlighted methods are where most of the time is spent by the* **new** *reference-counting system. `RecycleStagingInfo()` can now be compared with `VkRefCountManager.Decrement()`.*


TechnologicalPizza:
After some further review I will add unit tests for SparseDictionary and VkRefCountManager.

PR #416, opened 2021-11-30
Add support for creating swapchains on WinUI 3 based applications

ollitanska:
Setting xaml SwapChain panel has changed from UWP to WinUI 3. You now need to use interface Vortice.WinUI.ISwapChainPanelNative for it to work. 

This pull request updates Vortice to latest stable (as current version doens't have Vortice.WinUI package) and adds new function to create SwapChainSource on WinUI 3 projects.


shanet72:
What do I need to do to get this merged in?  I'd like to get Veldrid working under Maui, and not having these changes seems like the only blocker to that.

PR #417, opened 2021-12-04
Update ImageSharp to 1.0.4

awest03:
This fixes a bug where Veldrid.ImageSharp would not load the entirety of a texture (only the first few rows of pixels).


mellinoe:
@awest03 Thanks for the contribution!

mellinoe (approved):

PR #418, opened 2021-12-19
Support white-space characters in the file name of the mtllib row in obj files

NtFreX:
Currently the `ObjParser` throws an `ObjParseException` exception with the message `Expected exactly 1 components to a line starting with mtllib, on line 3, "mtllib Space Station Scene.mtl".` when the mtl file name contains spaces. The example exception above was thrown when using an exported file from Blender.

With this change everything following the `mtllib` key will be seen part of the filename.

I know this is conflicting with https://github.com/mellinoe/veldrid/pull/327

Feel free to throw this away if you don't need it. Thank you.

### DIFF
--- a/src/Veldrid.Utilities/ObjParser.cs
+++ b/src/Veldrid.Utilities/ObjParser.cs
@@ -147,8 +147,10 @@ namespace Veldrid.Utilities
                         ProcessFaceLine(pieces);
                         break;
                     case "mtllib":
-                        ExpectExactly(pieces, 1, "mtllib");
-                        DiscoverMaterialLib(pieces[1]);
+                        // file paths/file names can contain spaces, blender will not put quotation marks arround the file path. for example:
+                        // mtllib Space Station Scene.mtl
+                        string mtlLibFile = string.Join(" ", pieces.Skip(1));
+                        DiscoverMaterialLib(mtlLibFile);
                         break;
                     default:
                         throw new ObjParseException(


NtFreX:
I cannot find an official specification for the obj file and found conflicting information. So I don't know if you want to merge this.

-----------------------

http://paulbourke.net/dataformats/obj/

>mtllib filename1 filename2 . . .
>
>    Polygonal and free-form geometry statement.
>
>    Specifies the material library file for the material definitions
>    set with the usemtl statement. You can specify multiple filenames
>    with mtllib. If multiple filenames are specified, the first file
>    listed is searched first for the material definition, the second
>    file is searched next, and so on.
>
>    When you assign a material library using the Model program, only
>    one map library per .obj file is allowed. You can assign multiple
>    libraries using a text editor.
>
>    filename is the name of the library file that defines the
>    materials.  There is no default.

This seems to indicate that spaces are not supported and only one `mtllib` row is allowed.

--------------------

[Wikipedia](https://en.wikipedia.org/wiki/Wavefront_.obj_file)

> Materials that describe the visual aspects of the polygons are stored in external .mtl files. More than one external MTL material file may be referenced from within the OBJ file. The .mtl file may contain one or more named material definitions.
>
> mtllib [external .mtl file name]

This seems to indicate that spaces are supported and multiple `mtllib` rows are allowed.

------------------------

Blender just exports one row which can contain spaces.

mellinoe:
I'm still not sure that there is an official OBJ/MTL format specification. Since Blender is exporting files this way, and the fact that using multiple MTL files as a fallback seems like an extremely rare thing to do, this seems like a reasonable thing to support.

Can you add a test case that exercises this?

NtFreX:
I added some very simple tests. I am happy to add some more to it if needed.

PR #419, opened 2021-12-23
Fixes to support PBR sample

etinquis:
Details of the sample project can be found [here](https://github.com/mellinoe/veldrid-samples/pull/40).  There is also a related change in [veldrid-spirv](https://github.com/mellinoe/veldrid-spirv/pull/18) for support for setting shading lang version, though it shouldn't be a necessary prereq for any changes in this PR.

Some of the changes here include:
- Fixes for cubemap handling and some non-cubemap multi-layer cases across backends
- VkDescriptorPoolManager, vkGetImageMemoryRequirements2, and VK_KHR_portability_subset handling for running on MoltenVK without blowing up with validation errors
- D3D11 fix for binding samplers in compute shaders

Yet to do:
- [ ] Thorough regression check pass over test suite; mainly focused on new test cases up until now
  - [ ] Check OpenGLES (I've paid it no attention thus far)
- [ ] Address comments


src/Veldrid/Vk/VulkanUtil.cs
```diff
@@ -185,6 +185,20 @@ private static bool TryLoadVulkan()
                 srcStageFlags = VkPipelineStageFlags.Transfer;
                 dstStageFlags = VkPipelineStageFlags.FragmentShader;
             }
+            else if (oldLayout == VkImageLayout.General && newLayout == VkImageLayout.TransferDstOptimal)
```
etinquis, line 188: One of these blocks isn't necessary anymore; I noticed it's been fixed in master since.

src/Veldrid/Vk/VkTexture.cs
```diff
@@ -32,6 +32,7 @@ internal unsafe class VkTexture : Texture
         public override uint MipLevels { get; }
 
         public override uint ArrayLayers { get; }
+        public uint ActualArrayLayers => _actualImageArrayLayers;
```
etinquis, line 35: Not sure how best to handle this; I noticed there's some `effectiveArrayLayers` handling spread across a couple different places.  Is there somewhere common this can be moved that would be better for consolidation?

src/Veldrid/Vk/VkDeviceMemoryManager.cs
```diff
@@ -67,8 +67,30 @@ internal unsafe class VkDeviceMemoryManager : IDisposable
             VkImage dedicatedImage,
             Vulkan.VkBuffer dedicatedBuffer)
         {
-            // Round up to the nearest multiple of bufferImageGranularity.
-            size = ((size / _bufferImageGranularity) + 1) * _bufferImageGranularity;
+            if (dedicated)
```
etinquis, line 70: This was quarter-baked and I guarantee it's insufficient.  If anyone has suggestions on how this should be implemented, let me know.  Otherwise I'll have to dive in and try and figure it out from looking closer at the docs.

src/Veldrid/Vk/VkGraphicsDevice.cs
```diff
@@ -468,6 +468,11 @@ private void CreateInstance(bool debug, VulkanDeviceOptions options)
             StackList<IntPtr, Size64Bytes> instanceExtensions = new StackList<IntPtr, Size64Bytes>();
             StackList<IntPtr, Size64Bytes> instanceLayers = new StackList<IntPtr, Size64Bytes>();
 
+            if (availableInstanceExtensions.Contains(CommonStrings.VK_KHR_portability_subset))
```
etinquis, line 471: Not sure if this or the other similar change below is necessary.  Will need to double-check.

etinquis: (whichever one I did first didn't work alone)

src/Veldrid/Vk/VulkanUtil.cs
etinquis: Please lend your expertise as to what the flag values should be in the one that's still missing.  I tried to make a guess based on the rest of the cases, but I have no idea if it's correct.

src/Veldrid/OpenGL/OpenGLCommandExecutor.cs
```diff
@@ -941,14 +941,20 @@ public void SetComputeResourceSet(uint slot, ResourceSet rs, uint dynamicOffsetC
                         glTexViewRW.EnsureResourcesCreated();
                         if (pipeline.GetTextureBindingInfo(slot, element, out OpenGLTextureBindingSlotInfo imageBindingInfo))
                         {
+                            var layered = texViewRW.Target.Usage.HasFlag(TextureUsage.Cubemap) || texViewRW.ArrayLayers > 1;
+
+                            //todo: validate and test binding subsets of cubemap arrays
```
etinquis, line 946: I don't have a test for this yet.

src/Veldrid/OpenGL/OpenGLCommandExecutor.cs
```diff
@@ -941,14 +941,20 @@ public void SetComputeResourceSet(uint slot, ResourceSet rs, uint dynamicOffsetC
...
+                            if (layered && texViewRW.BaseArrayLayer > 0)
```
etinquis, line 947: Should probably throw if `texViewRW.ArrayLayers < ArrayLayers` as well.  I'm not sure what might be intended if someone were to try to create a view of a subset of layers (e.g. 2 faces of a cubemap) but I don't think that's supported by OpenGL here.

src/Veldrid/FormatHelpers.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Veldrid
 {
-    internal static class FormatHelpers
+    public static class FormatHelpers
```
etinquis, line 6: I found some code in the samples repo that duplicated the format->size conversion here for only one or two formats.   Since sizing and packing data is an important part of the binding process, I figured it'd be best to push shift the maintenance burden into the library and expose the canonical implementation of these conversions to anyone who needs to use it.

src/Veldrid.Tests/TestUtils.cs
```diff
@@ -229,7 +229,13 @@ protected Texture GetReadback(Texture texture)
                 Texture readback = RF.CreateTexture(ref desc);
                 CommandList cl = RF.CreateCommandList();
                 cl.Begin();
-                cl.CopyTexture(texture, readback);
+
+                //todo: temporary workaround due to cl.CopyTexture(texture, readback) not copying data properly for cubemaps in vulkan {
+                for (uint mip = 0; mip < texture.MipLevels; mip++)
```
etinquis, line 234: revert this; there was a fix in the Vulkan copy code that makes this workaround redundant.

src/Veldrid.Tests/RenderTests.cs
```diff
@@ -860,6 +858,70 @@ public void ComputeGeneratedTexture()
...
+                    var expectedColor = (byte)255.0f * ((t.layer + 1) * sideColorStep);
+                    var map = GD.Map<byte>(readback, MapMode.Read, subresource);
+
+                    Assert.All(
+                        from x in Enumerable.Range(0, (int)mipSize)
+                        from y in Enumerable.Range(0, (int)mipSize)
+                        select (X: x, Y: y),
+                        (xy) =>
+                        {
+                            Assert.InRange(map[xy.X, xy.Y], expectedColor - 1, expectedColor + 1);
```
etinquis, line 916: This is dumb.  It's working around undefined rounding rules coming back from the math in the shader.

I can probably pass values into the shader instead.  If you've got a better suggestion, let me know.

src/Veldrid.Tests/Veldrid.Tests.csproj
```diff
@@ -17,6 +17,7 @@
     <PackageReference Include="xunit" Version="2.3.1" />
     <PackageReference Include="xunit.console" Version="2.3.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
+    <PackageReference Include="Xunit.SkippableFact" Version="1.3.6" />
```
etinquis, line 20: It's somewhat dependent on how the CI system handles it, but I find that marking tests as skippable on conditions like `Skip.IfNot(GD.Features.ComputeShader);` helps to differentiate e.g. 'yes, compute shaders work for this back end' vs. 'couldn't validate one way or another whether or not compute shaders work for this back end' at a glance.

src/Veldrid.Tests/TextureTests.cs
```diff
@@ -135,6 +135,568 @@ public unsafe void Update_ThenCopySingleMip_Succeeds_R16UNorm()
...
+                for (uint face = 0; face < 6; face++)
+                {
+                    var mipSize = TexSize >> (int)mip;
+                    byte[] data = Enumerable.Repeat((face + 1) * 42, (int)(mipSize * mipSize)).Select(n => (byte)n).ToArray();
+                    GD.UpdateTexture(tex, data, 0, 0, 0, mipSize, mipSize, 1, mip, face);
+                }
+            }
+
+            Texture readback = GetReadback(tex);
+            Assert.All
```
etinquis, line 185: Though I found this to be the most pleasant assertion format to validate the data blocks in these tests, I still feel there must be a better way yet.

What I like about it is that it gives a better overview of the shape of the failure than simple loops because it aggregates the results instead of exiting after a single assertion failure.

src/Veldrid/Vk/VkGraphicsDevice.cs
mellinoe: This doesn't seem like it should be necessary, and I don't think this extension is finalized yet, so I'd rather not use it even optionally.

src/Veldrid/Vk/VkTexture.cs
mellinoe: I could see it moving into the base Texture class, but I'm not sure it would have the same kind of usefulness or meaning across all backends,or whether it would just introduce more confusion because of differences in cubemap/array layer handling.

src/Veldrid/Vk/VulkanUtil.cs
mellinoe: `VkImageLayout.General` should indicate that the resource was being used in a compute shader, so I'd expect srcStageFlags to be `VkPipelineStageFlags.Compute` and srcAccessmask to be something like `VkAccessFlags.ShaderWrite`

A lot of the transitions in here are... iffy. Really, this whole function should be refactored, but it's extremely tricky at this point.

src/Veldrid/Vk/VulkanUtil.cs
```diff
@@ -185,6 +185,20 @@ private static bool TryLoadVulkan()
...
+                barrier.srcAccessMask = VkAccessFlags.TransferRead;
+                barrier.dstAccessMask = VkAccessFlags.TransferWrite;
+                srcStageFlags = VkPipelineStageFlags.Transfer;
+                dstStageFlags = VkPipelineStageFlags.Transfer;
+            }
+            else if (oldLayout == VkImageLayout.General && newLayout == VkImageLayout.TransferSrcOptimal)
+            {
+                barrier.srcAccessMask = VkAccessFlags.TransferWrite;
+                barrier.dstAccessMask = VkAccessFlags.TransferRead;
+                srcStageFlags = VkPipelineStageFlags.Transfer;
```
mellinoe, line 199: VkPipelineStageFlags.~~Transfer~~Compute

src/Veldrid/Vk/VulkanUtil.cs
```diff
@@ -185,6 +185,20 @@ private static bool TryLoadVulkan()
...
+            else if (oldLayout == VkImageLayout.General && newLayout == VkImageLayout.TransferDstOptimal)
+            {
+                barrier.srcAccessMask = VkAccessFlags.TransferRead;
+                barrier.dstAccessMask = VkAccessFlags.TransferWrite;
+                srcStageFlags = VkPipelineStageFlags.Transfer;
+                dstStageFlags = VkPipelineStageFlags.Transfer;
+            }
+            else if (oldLayout == VkImageLayout.General && newLayout == VkImageLayout.TransferSrcOptimal)
+            {
+                barrier.srcAccessMask = VkAccessFlags.TransferWrite;
```
mellinoe, line 197: VkAccessFlags.~~TransferWrite~~ShaderWrite

mellinoe (changes requested):
Thanks for the great submission, and I'm very impressed with the sample work over in the other repository. Awesome stuff! I haven't had a chance to test this on my Macbook nor my Linux laptop yet, but there are a few test failures on my Windows machine with the current version. A few of them are with OpenGLES as you mentioned. Besides those test failures, I don't see any major problems here, just a few things to be cleaned up -- see other individual comments.

mellinoe:
By the way, is [this the issue](https://github.com/KhronosGroup/Vulkan-Portability/issues/14) you were seeing when running under MoltenVK? It seems like they've acknowledged that the new extension breaks everything, but haven't yet adjusted it (see [here](https://github.com/KhronosGroup/Vulkan-Portability/issues/18) too).

src/Veldrid/Vk/VulkanUtil.cs
etinquis: Updated, please take another look and confirm it's correct now.

src/Veldrid/FormatHelpers.cs
etinquis: Moved `GetSizeInBytes` to new public class `FormatSizeHelpers` and made `FormatHelpers` internal again.

src/Veldrid.Tests/TestUtils.cs
etinquis: Done

src/Veldrid.Tests/RenderTests.cs
etinquis: `floor` worked of course, thanks! :+1:

src/Veldrid.Tests/RenderTests.cs
```diff
@@ -860,6 +858,70 @@ public void ComputeGeneratedTexture()
...
+            GD.SubmitCommands(cl);
+            GD.WaitForIdle();
+
+            float sideColorStep = 1.0f / ArrayLayers;
+            Texture readback = GetReadback(computeOutput);
+            Assert.All
+            (
+                from mip in Enumerable.Range(0, (int)MipLevels)
+                from layer in Enumerable.Range(0, (int)ArrayLayers)
+                select (mip, layer),
```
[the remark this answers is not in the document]
etinquis: Swapped Assert.Alls with loops.  I have an urge to refactor all of them into some sort of additional Assert helper, but I'll save that for a different PR if I get around to it.

src/Veldrid.Tests/RenderTests.cs
```diff
@@ -860,6 +858,70 @@ public void ComputeGeneratedTexture()
...
+                ArrayLayers,
+                PixelFormat.R8_G8_B8_A8_UNorm,
+                TextureUsage.Sampled | TextureUsage.Storage);
+            Texture computeOutput = RF.CreateTexture(texDesc);
+            
+            ResourceLayout computeLayout = RF.CreateResourceLayout(new ResourceLayoutDescription(
+                new ResourceLayoutElementDescription("ComputeOutput", ResourceKind.TextureReadWrite, ShaderStages.Compute)));
+            ResourceSet computeSet = RF.CreateResourceSet(new ResourceSetDescription(computeLayout, computeOutput));
+
+            Pipeline computePipeline = RF.CreateComputePipeline(new ComputePipelineDescription(
```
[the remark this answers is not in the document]
etinquis: Veldrid.SPIRV v1.0.14 fixes it.  Updated in PR.

src/Veldrid.OpenGLBindings/OpenGLNative.cs
```diff
@@ -1884,6 +1885,12 @@ public static void LoadAllFunctions(IntPtr glContext, Func<string, IntPtr> getPr
                     LoadFunction("glCopyImageSubDataEXT", out p_glCopyImageSubData);
                 }
 
+                LoadFunction("glTextureView", out p_glTextureView);
+                if(p_glTextureView == null)
+                {
+                    LoadFunction("glTextureViewOES", out p_glTextureView);
```
etinquis, line 1891: I don't have a clean test for this.  The phone that I have that supports this extension ends up with the deadlocking issue described [here](https://community.khronos.org/t/adreno-510-gpu-deadlock-would-occur-when-trying-to-use-ubo-in-fragment-shader/106773/3).

src/Veldrid/OpenGL/OpenGLCommandExecutor.cs
```diff
@@ -964,14 +964,16 @@ public void SetComputeResourceSet(uint slot, ResourceSet rs, uint dynamicOffsetC
...
+                                    (uint)imageBindingInfo.RelativeIndex,
                                     glTexViewRW.Target.Texture,
                                     (int)texViewRW.BaseMipLevel,
                                     layered,
                                     (int)texViewRW.BaseArrayLayer,
                                     TextureAccess.ReadWrite,
                                     glTexViewRW.GetReadWriteSizedInternalFormat());
                                 CheckLastError();
+                                glUniform1i(imageBindingInfo.UniformLocation, imageBindingInfo.RelativeIndex);
+                                CheckLastError();
```
etinquis, line 976: [This commit](https://github.com/mellinoe/veldrid/commit/77588d4b6c1660be09ef2d4553610738f993a498#diff-c1c403b65fcd771b421112856cdf932bd732728f4a16267f99523ffbf6614bd8R631) appears to have split this implementation, but without this update my sample breaks.  Is there a way I can repro the problem that caused the divergence?

etinquis:
@mellinoe OpenGLES tests should be working now.

Strangely, on the one windows machine that I can get GLES initializing properly on, I'm seeing inconsistent test failures in Veldrid.Tests.OpenGLTextureTests.CubeMap_Copy_MultipleMip_CopySingleMipFaces that I don't see elsewhere.  Is that one of the ones that you were seeing fail earlier?

> By the way, is [this the issue](https://github.com/KhronosGroup/Vulkan-Portability/issues/14) you were seeing when running under MoltenVK? It seems like they've acknowledged that the new extension breaks everything, but haven't yet adjusted it (see [here](https://github.com/KhronosGroup/Vulkan-Portability/issues/18) too).

Yup, that looks like the one.

etinquis:
Nevermind, now I have GLES tests failing again.  Will look into it tomorrow.

src/Veldrid/Vk/VkDeviceMemoryManager.cs
TechnologicalPizza: Looking at the spec, it seems that sparse resource sizes must be a multiple of buffer-image granularity.
I am quite certain that dedicated resource sizes do not need to be a multiple of that granularity.

[In my fork](https://github.com/TechnologicalPizza/veldrid/blob/c65946a7e1caeece088e2433b5676940dd02e25b/src/Veldrid/Vk/VkDeviceMemoryManager.cs#L84) I've applied this as per spec and it works on Nvidia and AMD hardware.

I would also like to point out that the current alignment equation is flawed and will waste by one multiple when the value is `x + 1 = multiple` (look in my linked fork for a corrected equation).

TechnologicalPizza:
I can only add that GPU readback in unit tests is not fun. I somewhat managed to narrow it down to the unit test runner (running one test at a time has a higher success chance).
The tests could fail one day and work the other, and AMD especially didn't want to readback on OpenGL one day but worked after VS restart.
That was a horrible day...

mellinoe:
@etinquis Thanks again for the submission. I fixed up the OpenGL test failure -- there seemed to be a subtle driver issue with the DSA codepath when mapping a subset of a cubemap texture. The non-DSA codepath seemingly works fine, so I'm not inclined to dig into it that much. I also applied one of the suggestions from the conversation but otherwise took this as-is. I think there's more improvements to be made to the Texture API but this is a step in the right direction.

PR #420, opened 2022-01-02
Upload .nupkg's in GH workflow, remove AppVeyor entirely.

mellinoe:


zaafar:
If you don’t mind, may I suggest using https://github.com/marketplace/actions/publish-nuget it provide a couple of bells/whistles that Veldrid can use. :)

PR #424, opened 2022-01-05
Supporting the BlendFactor in the D3D backend

baSSiLL:
I wanted to use the `BlendStateDescription.BlendFactor` property along with `BlendFactor.BlendFactor` and `BlendFactor.InverseBlendFactor` enum values in an attachment. It turned out that this function is not implemented properly for the D3D backend. The value of the blend factor is not passed to `OMSetBlendState`, and also `BlendFactor.InverseBlendFactor` is converted to a wrong D3D value.
This pull request fixes the described issue and includes a test.

### DIFF
--- a/src/Veldrid.Tests/RenderTests.cs
+++ b/src/Veldrid.Tests/RenderTests.cs
@@ -1241,6 +1241,118 @@ namespace Veldrid.Tests
             }
             GD.Unmap(readback);
         }
+
+        [Fact]
+        public void UseBlendFactor()
+        {
+            const uint width = 512;
+            const uint height = 512;
+            using var output = RF.CreateTexture(
+                TextureDescription.Texture2D(width, height, 1, 1, PixelFormat.R32_G32_B32_A32_Float, TextureUsage.RenderTarget));
+            using var framebuffer = RF.CreateFramebuffer(new FramebufferDescription(null, output));
+
+            var yMod = GD.IsClipSpaceYInverted ? -1.0f : 1.0f;
+            var vertices = new[]
+            {
+                new ColoredVertex { Position = new Vector2(-1, 1 * yMod), Color = Vector4.One },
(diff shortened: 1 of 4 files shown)


baSSiLL:
As the `BlendStateDescription.BlendFactor` is not nullable, I decided to pass its value to `OMSetBlendState` always even though the corresponding parameter is nullable. Probably there's some reason to pass `null` when the blend factor is not used actually but I don't see it. If needed, this could be done by analyzing the `XXXFactor` properties in attachments for `BlendFactor.BlendFactor` and `BlendFactor.InverseBlendFactor` values.

baSSiLL:
Hi @mellinoe
I see the CI build fails with

> C:\Users\runneradmin\.nuget\packages\nerdbank.gitversioning\3.3.37\build\Nerdbank.GitVersioning.targets(115,5): error MSB4018: The "Nerdbank.GitVersioning.Tasks.SetCloudBuildVariables" task failed unexpectedly. [D:\a\veldrid\veldrid\src\Veldrid.ImageSharp\Veldrid.ImageSharp.csproj]
C:\Users\runneradmin\.nuget\packages\nerdbank.gitversioning\3.3.37\build\Nerdbank.GitVersioning.targets(115,5): error MSB4018: System.IO.IOException: The process cannot access the file 'D:\a\_temp\_runner_file_commands\set_env_2f27e3bc-4baf-4475-b4b1-06513a2afdd5' because it is being used by another process. [D:\a\veldrid\veldrid\src\Veldrid.ImageSharp\Veldrid.ImageSharp.csproj]

Which doesn't look caused by my commits. I guess this may be a temporary glitch which won't happen on the next build. Apparently I'm not the one who can force CI to rebuild :)

zaafar:
@baSSiLL push a dummy commit and it will re-trigger the CI

baSSiLL:
Thanks @zaafar I thought about this too but I don't like the idea to make dummy commits for any reason.
Besides, it won't help if there's some issue with the build environment. So it would be better that an owner took a look.

PR #429, opened 2022-01-17
update Nerdbank version

zaafar:
this fixes the issue where the Github CI pipeline crashes intermittently.


zaafar:
already done in this MR: https://github.com/mellinoe/veldrid/pull/432

PR #432, opened 2022-01-20
Update and improvements.

amerkoleci:
- Update Vortice.Windows to latest 2.1.0 stable release.
- Install .NET SDK 6.0 in CI build for future .NET 6.0 update.
- Cleanup dependencies.

### DIFF
--- a/src/NeoDemo/NeoDemo.csproj
+++ b/src/NeoDemo/NeoDemo.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net5.0</TargetFramework>
     <RootNamespace>Veldrid.NeoDemo</RootNamespace>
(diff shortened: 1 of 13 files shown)


amerkoleci:
Something was broken in Compute shader logic in Vortice.Direct3D11 thats causes tests to fail (like BasicCompute), seams something related to Map or CS binding logic (zzz)

src/NeoDemo/NeoDemo.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net5.0</TargetFramework>
```
TechnologicalPizza, line 5: I don't see why we would jump to net5 when net6 is the latest LTS.

amerkoleci: It was not part of this PR, but I can submit .NET 6.0 migration PR as well :D

TechnologicalPizza:
These failures may be caused by `BufferDescription.RawBuffer` being false by default.

PR #437, opened 2022-02-15
 Introduce TextureViewType and handle detection from Texture

### LINKED ISSUE #245: Vulkan: Cannot explicitly create texture arrays
Creating texture arrays with 1 layer is impossible, as implemented [here](https://github.com/mellinoe/veldrid/blob/0be2107806d3f3543132fdfdead2443fee1713a3/src/Veldrid/Vk/VkTextureView.cs#L60)

amerkoleci:
- Introduce TextureViewType and handle detection from Texture, fixes #245.
- Support for all backends.

### DIFF
--- a/src/Veldrid.Tests/TextureTests.cs
+++ b/src/Veldrid.Tests/TextureTests.cs
@@ -935,6 +935,16 @@ namespace Veldrid.Tests
             Assert.Equal(1u, tex.Height);
         }
 
+        [Fact]
+        public void CreateTextureView2DArray()
+        {
+            Texture texture = RF.CreateTexture(TextureDescription.Texture2D(1, 1, 1, 1, PixelFormat.R8_G8_B8_A8_UNorm, TextureUsage.Sampled));
+            TextureView view = RF.CreateTextureView(texture, TextureViewType.View2DArray);
+            Assert.Equal(TextureType.Texture2D, texture.Type);
+            Assert.Equal(TextureViewType.View2DArray, view.ViewType);
+        }
+
+
         private static readonly FormatProps[] s_allFormatProps =
         {
             new FormatProps(PixelFormat.R8_UNorm, 8, 0, 0, 0),
--- a/src/Veldrid/D3D11/D3D11TextureView.cs
+++ b/src/Veldrid/D3D11/D3D11TextureView.cs
@@ -5,7 +5,7 @@ namespace Veldrid.D3D11
 {
     internal class D3D11TextureView : TextureView
     {
-        private string _name;
+        private string? _name;
         private bool _disposed;
 
         public ID3D11ShaderResourceView ShaderResourceView { get; }
@@ -18,6 +18,7 @@ namespace Veldrid.D3D11
             D3D11Texture d3dTex = Util.AssertSubtype<Texture, D3D11Texture>(description.Target);
             ShaderResourceViewDescription srvDesc = D3D11Util.GetSrvDesc(
                 d3dTex,
+                description.ViewType,
                 description.BaseMipLevel,
                 description.MipLevels,
                 description.BaseArrayLayer,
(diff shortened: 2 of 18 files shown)


src/Veldrid/TextureViewDescription.cs
```diff
@@ -102,14 +121,54 @@ public TextureViewDescription(Texture target, uint baseMipLevel, uint mipLevels,
         /// <param name="mipLevels">The number of mip levels visible in the view.</param>
         /// <param name="baseArrayLayer">The base array layer visible in the view.</param>
         /// <param name="arrayLayers">The number of array layers visible in the view.</param>
-        public TextureViewDescription(Texture target, PixelFormat format, uint baseMipLevel, uint mipLevels, uint baseArrayLayer, uint arrayLayers)
+        /// <param name="viewType">The view type to create or <see cref="TextureViewType.Undefined"/> to detect from texture.</param>
+        public TextureViewDescription(Texture target, PixelFormat format, uint baseMipLevel, uint mipLevels, uint baseArrayLayer, uint arrayLayers, TextureViewType viewType = TextureViewType.Undefined)
         {
             Target = target;
+            if (viewType == TextureViewType.Undefined)
```
mellinoe, line 128: Please surrround with braces even if it's a one-line block

src/Veldrid/D3D11/D3D11TextureView.cs
```diff
@@ -30,58 +31,47 @@ public D3D11TextureView(D3D11GraphicsDevice gd, ref TextureViewDescription descr
...
+                        uavDesc.Texture3D.WSize = (int)d3dTex.Depth;
+                        break;
                 }
 
                 UnorderedAccessView = device.CreateUnorderedAccessView(d3dTex.DeviceTexture, uavDesc);
             }
         }
 
-        public override string Name
+        public override string? Name
```
mellinoe, line 74: This doesn't really seem to be related, although I am supportive of trying to correctly annotate nullability in the library over time. Could you remove it from this PR, though?

src/Veldrid/TextureViewDescription.cs
```diff
@@ -41,9 +45,14 @@ public struct TextureViewDescription : IEquatable<TextureViewDescription>
         /// </summary>
         /// <param name="target">The desired target <see cref="Texture"/>. This <see cref="Texture"/> must have been created
         /// with the <see cref="TextureUsage.Sampled"/> usage flag.</param>
-        public TextureViewDescription(Texture target)
+        /// <param name="viewType">The view type to create or <see cref="TextureViewType.Undefined"/> to detect from texture.</param>
+        public TextureViewDescription(Texture target, TextureViewType viewType = TextureViewType.Undefined)
         {
             Target = target;
+            if (viewType == TextureViewType.Undefined)
```
mellinoe, line 52: Same here, please surround with braces

src/Veldrid/TextureViewDescription.cs
```diff
@@ -102,14 +121,54 @@ public TextureViewDescription(Texture target, uint baseMipLevel, uint mipLevels,
...
+                    if (target.ArrayLayers > 1)
+                    {
+                        return TextureViewType.View1DArray;
+                    }
+
+                    return TextureViewType.View1D;
+                case TextureType.Texture2D:
+                    if ((usage & TextureUsage.Cubemap) == TextureUsage.Cubemap)
+                    {
+                        if (target.ArrayLayers > 1)
```
mellinoe, line 153: nit: ternary expression would be better here (and is used elsewhere in related sections), e.g.

```C#
return target.ArrayLayers > 1 ? TextureViewType.ViewCubeArray : TextureViewType.ViewCube;
```

Same for the next section (View2DArray vs. View2D)

mellinoe (commented):
Looks pretty good overall. Could you please add at least one test case covering the most important use for this, e.g. using a non-array Texture inside an array view? It also needs to have a full implementation for all backends, but I think you're already closing the gap there.

src/Veldrid/TextureViewDescription.cs
amerkoleci: Done

src/Veldrid/D3D11/D3D11TextureView.cs
amerkoleci: Done

src/Veldrid/TextureViewDescription.cs
amerkoleci: Done

amerkoleci:
> Looks pretty good overall. Could you please add at least one test case covering the most important use for this, e.g. using a non-array Texture inside an array view? It also needs to have a full implementation for all backends, but I think you're already closing the gap there.

Test just landed!

src/Veldrid/TextureViewDescription.cs
```diff
@@ -41,9 +45,11 @@ public struct TextureViewDescription : IEquatable<TextureViewDescription>
         /// </summary>
         /// <param name="target">The desired target <see cref="Texture"/>. This <see cref="Texture"/> must have been created
         /// with the <see cref="TextureUsage.Sampled"/> usage flag.</param>
-        public TextureViewDescription(Texture target)
+        /// <param name="viewType">The view type to create or <see cref="TextureViewType.Undefined"/> to detect from texture.</param>
+        public TextureViewDescription(Texture target, TextureViewType viewType = TextureViewType.Undefined)
```
mellinoe, line 49: All of these constructor changes need to be reverted, because adding a new parameter (even if it has a default value) is not a backwards-compatible change for the ABI. Old code would have to be recompiled to work with these new signatures.

In order to add new a new field like this, you should add one single new constructor that accepts all of the other fields, plus the new one. This is generally why the set of constructors for Veldrid types looks the way it does -- new values have been added over time and new constructors with them.

src/Veldrid.Tests/TextureTests.cs
```diff
@@ -935,6 +935,16 @@ public void CreateSmallTexture(PixelFormat format)
             Assert.Equal(1u, tex.Height);
         }
 
+        [Fact]
+        public void CreateTextureView2DArray()
```
mellinoe, line 939: We need a test that functionally stresses the new codepaths that will be enabled by this new feature. This test doesn't really do anything except check that the constructors have assigned reasonable values to stuff. That is valuable by itself, but the hardest part of this feature is making sure everything behaves correctly and identically across backends. A test that does something like the following would give a lot more confidence:

1. Create an array texture and fill it with some color, different on each layer
2. Create a NON-array view of a specific layer
3. Sample from the texture
4. Assert that the retrieved color is correct depending on the layer

and/or

1. Create a NON-array texture and fill it with some color
2. Create a 2D-Array view of it
3. Sample from the texture view at a certain layer
4. Assert that the correct color was retrieved

src/Veldrid/TextureViewDescription.cs
amerkoleci: Done!

src/Veldrid.Tests/TextureTests.cs
amerkoleci: I agree, but I won't be able to do that, feel  free to close PR is it makes more breaking change.

PR #439, opened 2022-02-19
Add "color write" masking support in blend attachments

frenzibyte:
I was relying on `BlendFactor` at first, but turns out that doesn't really play well with OpenGL. So I looked at adding support for color write masks instead, which exists in all graphics APIs and is fairly simple to implement.

### DIFF
--- a/src/Veldrid/BlendAttachmentDescription.cs
+++ b/src/Veldrid/BlendAttachmentDescription.cs
@@ -12,6 +12,10 @@ namespace Veldrid
         /// </summary>
         public bool BlendEnabled;
         /// <summary>
+        /// Controls which components of the color will be written to the framebuffer.
+        /// </summary>
+        public ColorWriteMask ColorWriteMask;
+        /// <summary>
         /// Controls the source color's influence on the blend result.
         /// </summary>
         public BlendFactor SourceColorFactor;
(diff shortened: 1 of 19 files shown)


src/Veldrid/BlendAttachmentDescription.cs
```diff
@@ -12,6 +12,10 @@ public struct BlendAttachmentDescription : IEquatable<BlendAttachmentDescription
         /// </summary>
         public bool BlendEnabled;
         /// <summary>
+        /// Controls which components of the color will be written to the framebuffer.
+        /// </summary>
+        public ColorWriteMask ColorWriteMask;
```
mellinoe, line 17: The only concern I have is that if you have a default-constructed BlendAttachmentDescription, then this is a breaking behavioral change, and a big one at that. E.g. if I am already doing this:

```
BlendAttachmentDescription desc = default();
desc.... = ...;
// etc.
UseDesc(desc);
```

I will now be using an empty ColorWriteMask and nothing will work anymore.

It's ugly, but in the name of backwards compatibility we need to make this a nullable field (`ColorWriteMask?`) so that default-constructed objects still work correctly. Nothing else about the constructors you modified need to be adjusted (they can still assign a non-null value to the field, as long as the behavior is the same as before).

Then, I'd introduce a helper function that converts from a `ColorWriteMask?` to `ColorWriteMask`, understanding that a `null` mask is the same as a mask with `.All` assigned, and use that in all the other places that are reading from the BlendAttachmentDescription field. Everything else would stay the same.

mellinoe (changes requested):
Looks really good overall. The abstraction is sound and maps to all the backends really cleanly. I made one comment about backwards compatibility which needs to be addressed given the nature of the change.

I'd also really like to see a test case for this, if it's feasible, or integration into NeoDemo in some way -- some way to functionally validate this.

1. Unit test -- Copy-paste one of the test cases in RenderTests and have it apply a ColorWriteMask before rendering stuff, then modify the assertions based on the mask being used (e.g. only red was written, only green, etc.).
2. NeoDemo already has a "Tinted output" checkbox in the "Settings" menu, which uses a hard-coded shader to apply a red tint. It would be great if this feature was used instead, and even better would be to provide the user a color selector widget in the Settings menu so that they could control the write mask used (instead of always being red).

src/Veldrid/BlendAttachmentDescription.cs
frenzibyte: Good note, have made it nullable as requested. I'm not familiar with the helper patterns in Veldrid, but I've added a public extension method (`GetOrDefault`) for it. Should it be internal instead? or a non-extension static method?

frenzibyte:
> I'd also really like to see a test case for this, if it's feasible, or integration into NeoDemo in some way -- some way to functionally validate this.
>
> Unit test -- Copy-paste one of the test cases in RenderTests and have it apply a ColorWriteMask before rendering stuff, then modify the assertions based on the mask being used (e.g. only red was written, only green, etc.).
NeoDemo already has a "Tinted output" checkbox in the "Settings" menu, which uses a hard-coded shader to apply a red tint. It would be great if this feature was used instead, and even better would be to provide the user a color selector widget in the Settings menu so that they could control the write mask used (instead of always being red).

Right, I've added these now as requested. I've added a "Color mask" menu containing checkboxes for each color component, to allow for in-between color masks:

https://user-images.githubusercontent.com/22781491/166129909-1e8bf011-490a-4561-9169-e5bfb90c74d2.mp4

I've also kept the "Tinted output" checkbox to keep the feature of testing multiple render targets covered.

PR #442, opened 2022-03-11
fix IdxOffset field

bvigario:
https://github.com/ocornut/imgui/issues/4845


mellinoe:
Looks good to me, thank you for the contribution!

PR #445, opened 2022-03-20
Update ImageSharp from 1.0.4 to 2.1.0

NtFreX:
The method `TryGetSinglePixelSpan` has been replaced with pixel processors to ensure memory safety while increasing perf.
https://github.com/SixLabors/ImageSharp/issues/1739

`DangerousTryGetSinglePixelMemory` is an alternative to pixel processors recommended for interop functionality.

I did not test the code yet and will add some if I find the time and ability.

### DIFF
--- a/src/Veldrid.ImageSharp/ImageSharpCubemapTexture.cs
+++ b/src/Veldrid.ImageSharp/ImageSharpCubemapTexture.cs
@@ -1,9 +1,8 @@
 ﻿using SixLabors.ImageSharp;
-using SixLabors.ImageSharp.Advanced;
 using SixLabors.ImageSharp.PixelFormats;
 using System;
+using System.Buffers;
 using System.IO;
-using System.Runtime.InteropServices;
 
 namespace Veldrid.ImageSharp
 {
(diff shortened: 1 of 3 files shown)


Thraka:
@NtFreX I believe this puts the users of the library into possible commercial licensing if they sell they're resulting apps. This new licensing model (I think) started with v2 of image sharp. By keeping v1 of image sharp, you still get the old licensing model which is 100% free.

https://sixlabors.com/pricing/
https://sixlabors.com/pricing/license/

NtFreX:
@Thraka Oh yeah makes sense. There isn't a new feature that is being used in this pr so this would only have been for ~perf/maintainability. I guess this can be deleted.

Saalvage:
In the case that anyone else stumbles across this: The reason why `DangerousTryGetSinglePixelMemory` fails in the demo is because `PreferContiguousImageBuffers` has not been set in the `Configuration` used by the `Image.Load` and `Image.Clone` methods. While the library itself should pass this configuration when calling the methods, the provided fallback should still exist for the constructors taking in a user-provided `Image<Rgba32>`.

PR #447, opened 2022-04-12
Use constant stackalloc in SetObjectLabel

SupinePandora43:
Fix possible StackOverflowException (in case of a crazy opengl driver)
Fix `""` (empty) label causing crash (because `stackalloc byte[0]` returns `(byte*) null`
Improve performance by using a constant-sized stackalloc
Use null-terminated string


mellinoe:
Looks reasonable, thanks

PR #450, opened 2022-05-02
[Vk] Increment resource refs immediately

TechnologicalPizza:


mellinoe:
Previously, ref counts were only incremented when the CommandList was actually submitted. Now, they are incremented when a command is recorded, but they still aren't decremented until the command completes. However, there's no guarantee that a user will submit the CommandList even after they've recorded it. This will leak references to the resources that were used in it and they will never be released, `Dispose` doesn't do it.

TechnologicalPizza:
Excellent point. Simply decrementing all refs upon disposal should be enough right?

mellinoe:
Something like that, but it may be more complicated because `_currentStagingInfo` is recreated / recycled in some cases (I'd need to look through all the codepaths to remember how it's treated). Basically, any time `_currentStagingInfo` is thrown away but still has some refcounts in it, then they need to be released. That might only be in the Dispose method, but I'm not sure.

TechnologicalPizza:
Looks like `_currentStagingInfo` is only released in `CommandList.Begin`. Decrementing refs when disposing should be trivial.

PR #452, opened 2022-05-06
Use dated version suffix

SupinePandora43:
To make myget packages versionable


mellinoe:
I don't think this is cooperating with `Nerdbank.GitVersioning` because when I run it locally it has no effect. Unfortunately the GH Actions workflow is set to skip even building the NuGet package unless it's an official build so I can't see how it works there. Could you adjust the workflow so that it builds the package and just doesn't upload it for CI jobs?

PR #453, opened 2022-05-31
Name all resources created in ImGuiRenderer

Vectron:
This makes it easier to identify resources belonging to the imgui rendering in renderdoc


mellinoe:
Looks reasonable to me. Thanks!

PR #454, opened 2022-06-06
Add validation of GL structured buffer names, fix warnings due to stackalloc inside loops.

csinkers:
Fixed a few instances of CA2014 ("Do not use stackalloc in loops"), and also added better exception messages after trying to use an invalid structured buffer name when compiling with GL_VALIDATE_SHADER_RESOURCE_NAMES.

### DIFF
--- a/src/Veldrid.OpenGLBindings/OpenGLNative.cs
+++ b/src/Veldrid.OpenGLBindings/OpenGLNative.cs
@@ -919,10 +919,25 @@ namespace Veldrid.OpenGLBinding
             => p_glDispatchCompute(num_groups_x, num_groups_y, num_groups_z);
 
         [UnmanagedFunctionPointer(CallConv)]
+        private delegate uint glGetProgramInterfaceiv_t(uint program, ProgramInterface programInterface, ProgramInterfaceParameterName pname, int* @params);
+        private static glGetProgramInterfaceiv_t p_glGetProgramInterfaceiv;
+        public static uint glGetProgramInterfaceiv(uint program, ProgramInterface programInterface, ProgramInterfaceParameterName pname, int* @params)
+            => p_glGetProgramInterfaceiv(program, programInterface, pname, @params);
+
+        [UnmanagedFunctionPointer(CallConv)]
         private delegate uint glGetProgramResourceIndex_t(uint program, ProgramInterface programInterface, byte* name);
         private static glGetProgramResourceIndex_t p_glGetProgramResourceIndex;
         public static uint glGetProgramResourceIndex(uint program, ProgramInterface programInterface, byte* name)
             => p_glGetProgramResourceIndex(program, programInterface, name);
+
+        [UnmanagedFunctionPointer(CallConv)]
+        private delegate uint glGetProgramResourceName_t(uint program, ProgramInterface programInterface, uint index, uint bufSize, uint* length, byte* name);
+        private static glGetProgramResourceName_t p_glGetProgramResourceName;
+        public static uint glGetProgramResourceName(uint program, ProgramInterface programInterface, uint index, uint bufSize, uint* length, byte* name)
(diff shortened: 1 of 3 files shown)


src/Veldrid.OpenGLBindings/OpenGLNative.cs
```diff
@@ -918,12 +918,27 @@ public static void glObjectLabel(ObjectLabelIdentifier identifier, uint name, ui
...
         public static uint glGetProgramResourceIndex(uint program, ProgramInterface programInterface, byte* name)
             => p_glGetProgramResourceIndex(program, programInterface, name);
 
+        [UnmanagedFunctionPointer(CallConv)]
+        private delegate uint glGetProgramResourceName_t(uint program, ProgramInterface programInterface, uint index, uint bufSize, uint* length, byte* name);
+        private static glGetProgramResourceName_t p_glGetProgramResourceName;
+        public static uint glGetProgramResourceName(uint program, ProgramInterface programInterface, uint index, uint bufSize, uint* length, byte* name)
+            => p_glGetProgramResourceName(program, programInterface, index, bufSize, length, name);
+
+        // LoadFunction("glGetProgramInterface", out p_glGetProgramInterface);
```
mellinoe, line 939: Remove

src/Veldrid/OpenGL/OpenGLPipeline.cs
```diff
@@ -148,21 +136,10 @@ private void CreateGraphicsGLResources()
...
-                    {
-                        int bytesWritten = Encoding.UTF8.GetBytes(charPtr, elementName.Length, elementNamePtr, byteCount);
-                        Debug.Assert(bytesWritten == byteCount - 1);
-                    }
-                    elementNamePtr[byteCount - 1] = 0; // Add null terminator.
-
-                    int location = glGetAttribLocation(_program, elementNamePtr);
+                    int location = GetAttribLocation(layoutDesc.Elements[i].Name);
                     if (location == -1)
-                    {
```
mellinoe, line 163: Prefer to keep these blocks even if they are one-line in length.

src/Veldrid/OpenGL/OpenGLPipeline.cs
```diff
@@ -364,7 +377,31 @@ void ReportInvalidResourceName(string resourceName)
...
 
-            throw new VeldridException($"Unable to bind uniform \"{resourceName}\" by name. Valid names for this pipeline are: {string.Join(", ", names)}");
+            throw new VeldridException($"Unable to bind uniform \"{uniformName}\" by name. Valid names for this pipeline are: {string.Join(", ", names)}");
+        }
+
+        void ReportInvalidResourceName(string resourceName, ProgramInterface resourceType)
+        {
+            int maxLength = 0;
+            int resourceCount = 0;
+            glGetProgramInterfaceiv(_program, resourceType, ProgramInterfaceParameterName.MaxNameLength, &maxLength);
```
mellinoe, line 387: These need to be hidden behind a capability/version check, because they are only core in version 4.3+, even if they are behind the DEBUG and GL_VALIDATE_SHADER_RESOURCE_NAMES flags

mellinoe (commented):
This looks good overall, this will definitely help diagnose buffer naming issues which is a common problem.

src/Veldrid/OpenGL/OpenGLPipeline.cs
csinkers: @mellinoe 
Updated - is this the preferred pattern for version checks?

mellinoe: It would be preferable to provide a fallback for version that don't support this particular functionality. Is that feasible in this case?

mellinoe (approved):
Looks good to me. Thanks for the improvement!

PR #459, opened 2022-08-26
Added an exception throw when a ResourceLayout slot is null

DiegoG1019:
When a ResourceLayout array was bound to a Pipeline and, by accident, one of its members remained `null`, the CommandList would throw a `NullReferenceException` that described very little of the problem and therefore was hard to figure out and debug.

This pull request fixes the issue by adding a check through a null-coalescing operator

### DIFF
--- a/src/Veldrid/CommandList.cs
+++ b/src/Veldrid/CommandList.cs
@@ -222,7 +222,9 @@ namespace Veldrid
                     $"Failed to bind ResourceSet to slot {slot}. The active graphics Pipeline only contains {layoutsCount} ResourceLayouts.");
             }
 
-            ResourceLayout layout = _graphicsPipeline.ResourceLayouts[slot];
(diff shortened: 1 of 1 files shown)


src/Veldrid/ValidationHelpers.cs
```diff
@@ -8,6 +9,8 @@ internal static class ValidationHelpers
         internal static void ValidateResourceSet(GraphicsDevice gd, ref ResourceSetDescription description)
         {
 #if VALIDATE_USAGE
+            if (description.Layout is null)
+                throw new NullReferenceException("There is no associated ResourceLayout in the provided ResourceSetDescription");
```
mellinoe, line 13: `NullReferenceException` isn't right here, I would just make this a `VeldridException`. Veldrid doesn't really throw a large variety of different exceptions.

PR #462, opened 2022-10-21
Added Vulkan support for copying depth textures

aquilinum0:
The Vulkan backend currently does not support copying depth textures.

### DIFF
--- a/src/Veldrid/Vk/VkCommandList.cs
+++ b/src/Veldrid/Vk/VkCommandList.cs
@@ -815,17 +815,23 @@ namespace Veldrid.Vk
 
             if (!sourceIsStaging && !destIsStaging)
             {
+                var srcAspect = (srcVkTexture.Usage & TextureUsage.DepthStencil) != 0
+                    ? VkImageAspectFlags.Depth
+                    : VkImageAspectFlags.Color;
                 VkImageSubresourceLayers srcSubresource = new VkImageSubresourceLayers
                 {
-                    aspectMask = VkImageAspectFlags.Color,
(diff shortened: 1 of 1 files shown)


src/Veldrid/Vk/VkCommandList.cs
```diff
@@ -815,17 +815,23 @@ private protected override void UpdateBufferCore(DeviceBuffer buffer, uint buffe
...
                 VkImageSubresourceLayers srcSubresource = new VkImageSubresourceLayers
                 {
-                    aspectMask = VkImageAspectFlags.Color,
+                    aspectMask = srcAspect,
                     layerCount = layerCount,
                     mipLevel = srcMipLevel,
                     baseArrayLayer = srcBaseArrayLayer
                 };
 
+                var dstAspect = (dstVkTexture.Usage & TextureUsage.DepthStencil) != 0
```
mellinoe, line 829: I only question whether `VkImageAspectFlags.Stencil` is permitted to be passed here if the texture has a depth-only format. There is at least one other place in the code that treats it separately: https://github.com/mellinoe/veldrid/blob/eef8375169d1960a322f47f95e9b6ee8126e7b43/src/Veldrid/Vk/VkFramebuffer.cs#L202. I suspect a validation error might be emitted if you use the Stencil flag but the texture is depth-only.

This would be confirmed one way or the other by adding a test that exercises this.

mellinoe (changes requested):
Looks correct overall, but there should be a test case that exercises this. Presumably this precise codepath isn't covered as the tests are all passing currently.

PR #466, opened 2022-10-27
Update to Vortice 2.1.32 to fix crash on M360 Intel Integrated Graphics

seanspicer:
Eric, 

This pull request is needed to fix a crash that is currently occurring on Intel Integrated Graphics (M360 specifically).   The crash occurs due to Vortice 2.1.0 incorrectly setting a uniform buffer pointer to IntPtr.Zero in VSUnsetConstantBuffer (ID3D11DeviceContext.cs).   Recent changes have corrected the issue.


mellinoe:
Looks good to me; thanks for the fix.

Issue #456, opened 2022-07-16
Android .NET 6.0 CreateOpenGLES() failed to get the default Android EGLDisplay


eightonegulf:
Hi,
I'm trying to setup Veldrid within an Android .NET 6.0 application.
My MainActivity class looks as follow:

```
    public class MainActivity : Activity
    {
        protected override void OnCreate(Bundle? savedInstanceState)
        {
            base.OnCreate(savedInstanceState);
            GraphicsDeviceOptions options = new GraphicsDeviceOptions(
                false,
                PixelFormat.R16_UNorm,
                false,
                ResourceBindingModel.Improved,
                true,
                true);
            var surface = new VeldridSurfaceView(this, Veldrid.GraphicsBackend.OpenGLES, options);
            SetContentView(surface);
        }
    }
```

The VeldridSurfaceView is directly copied from
https://github.com/mellinoe/veldrid-samples/blob/master/src/SampleBase.Android/VeldridSurfaceView.cs

When I run this on my Android device, line #84 in VeldridSurfaceView `GraphicsDevice = GraphicsDevice.CreateOpenGLES(DeviceOptions, sd);` will throw an exception stating;

CreateOpenGLES() failed to get the default Android EGLDisplay: BadParameter

This exception originates from https://github.com/mellinoe/veldrid/blob/master/src/Veldrid/OpenGL/OpenGLGraphicsDevice.cs
InitializeANativeWindow(), line #621.
For some reason the display can't be found.

When I run exactly the same code from an Xamarin Android app, everything works as expected, and I can use Veldrid just fine.
Only when I run from a .NET 6.0 app this goes wrong.

I am using VS 2022 Preview in order to use .NET 6.0 on Android.

I can't figure out what I'm missing. Is this a compatability issue, or am I missing something else?

Thanks.

0x5b25:
Hi, that might be caused by the wrong `eglGetDisplay` signature in `EGLNative.cs`, the actual c function requires a argument of pointer type , but the C# wrapper supplies a System.Int32 type, on 64-bit Android OS (which is the majority now) this causes a invalid argument exception because the upper 32 bit half is just rubbish data.

mellinoe:
It would be great to confirm if @0x5b25 's change does indeed fix it. I don't have an active Android setup right now so this will take a bit for me to debug.

xtuzy:
Use IntPtr fixed it
[...]

PR #473, opened 2022-12-21
Updated imgui.net to 1.89.1

zaafar:


src/Veldrid.ImGui/ImGuiRenderer.cs
```diff
@@ -452,19 +452,19 @@ ImGuiKey keyToImGuiKeyShortcut(Key keyToConvert, Key startKey1, ImGuiKey startKe
             {
                 case Key.ShiftLeft:
                 case Key.ShiftRight:
-                    result = ImGuiKey.ModShift;
+                    result = ImGuiKey.ImGuiMod_Shift;
```
mellinoe, line 455: This is kind of awkward and might warrant a better translation over in ImGui.NET.

zaafar: Feel free to come up with a better strat. Once it’s merged I’ll update this PR and 4ish others. :)

zaafar: FYI:

Renamed and merged keyboard modifiers key enums and flags into a same set:  (#4921, #456)
   - ImGuiKey_ModCtrl  and ImGuiModFlags_Ctrl  -> ImGuiMod_Ctrl
   - ImGuiKey_ModShift and ImGuiModFlags_Shift -> ImGuiMod_Shift
   - ImGuiKey_ModAlt   and ImGuiModFlags_Alt   -> ImGuiMod_Alt
   - ImGuiKey_ModSuper and ImGuiModFlags_Super -> ImGuiMod_Super
  Kept inline redirection enums (will obsolete).
  This change simplifies a few things, reduces confusion, and will facilitate upcoming
  shortcut/input ownership apis.
  - The ImGuiKey_ModXXX were introduced in 1.87 and mostly used by backends.
  - The ImGuiModFlags_XXX have been exposed in imgui.h but not really used by any public api,
    only by third-party extensions. They were however subject to a recent rename
    (ImGuiKeyModFlags_XXX -> ImGuiModFlags_XXX) and we are exceptionally commenting out
    the older ImGuiKeyModFlags_XXX names ahead of obsolescence schedule to reduce confusion
    and because they were not meant to be used anyway.

mellinoe:
I went ahead and updated to v1.89.2, which includes a change to the generator so that ImGuiKey retains backwards compatibility.

PR #474, opened 2022-12-22
Update NativeLibraryLoader to v1.0.13

trolleyman:
Thanks for the merge in NativeLibraryLoader @mellinoe :) - this is the MR for Veldrid to update it to use the new code.


mellinoe:
LGTM, thanks for the fix here and over in the NativeLibraryLoader repo.

PR #483, opened 2023-03-21
Add support for feature level 10_0

smoogipoo:
# Do not merge

I'm making this branch both as a tracking branch for our own fork where we use feature level 10_0 to target older devices like [9800gt](https://www.techpowerup.com/gpu-specs/geforce-9800-gt.c635), and as a place to discuss how this could be mainlined if at all.

For example, should I track the selected feature level to retarget the shader version?

### DIFF
--- a/src/Veldrid/D3D11/D3D11Shader.cs
+++ b/src/Veldrid/D3D11/D3D11Shader.cs
@@ -60,22 +60,22 @@ namespace Veldrid.D3D11
             switch (description.Stage)
             {
                 case ShaderStages.Vertex:
-                    profile = "vs_5_0";
+                    profile = "vs_4_0";
                     break;
                 case ShaderStages.Geometry:
-                    profile = "gs_5_0";
+                    profile = "gs_4_0";
                     break;
                 case ShaderStages.TessellationControl:
-                    profile = "hs_5_0";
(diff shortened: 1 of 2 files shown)


src/Veldrid/D3D11/D3D11Shader.cs
```diff
@@ -60,22 +60,22 @@ private byte[] CompileCode(ShaderDescription description)
...
-                    profile = "ds_5_0";
+                    profile = "ds_4_0";
                     break;
                 case ShaderStages.Fragment:
-                    profile = "ps_5_0";
+                    profile = "ps_4_0";
                     break;
                 case ShaderStages.Compute:
-                    profile = "cs_5_0";
+                    profile = "cs_4_0";
```
frenzibyte, line 78: A major limitation in `cs_4_0` is the inability to bind `RWTexture2D` UAVs (known as "typed UAVs", although `RWStructuredBuffer` is still supported), and the inability to bind more than one UAV resource. This should be handled somewhere in Veldrid (probably as a feature in `GraphicsDeviceFeatures`).

See https://learn.microsoft.com/en-us/windows/win32/direct3d11/direct3d-11-advanced-stages-compute-shader#using-compute-shader-on-direct3d-10x-hardware for more information.

PR #484, opened 2023-03-24
Use flip model for D3D11 swapchains

smoogipoo:
As documented [here](https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/dxgi-flip-model), [here](https://www.youtube.com/watch?v=E3wTajGZOsA) and [here](https://github.com/MicrosoftDocs/win32/blob/docs/desktop-src/direct3ddxgi/for-best-performance--use-dxgi-flip-model.md), the flip model has performance advantages compared to the blt model.

### DIFF
--- a/src/Veldrid/D3D11/D3D11Swapchain.cs
+++ b/src/Veldrid/D3D11/D3D11Swapchain.cs
@@ -88,7 +88,7 @@ namespace Veldrid.D3D11
                         (int)description.Width, (int)description.Height, _colorFormat),
                     OutputWindow = win32Source.Hwnd,
                     SampleDescription = new SampleDescription(1, 0),
-                    SwapEffect = SwapEffect.Discard,
+                    SwapEffect = SwapEffect.FlipDiscard,
                     BufferUsage = Usage.RenderTargetOutput
                 };
 


src/Veldrid/D3D11/D3D11Util.cs
```diff
@@ -1,9 +1,20 @@
...
     internal static class D3D11Util
     {
+        /// <summary>
+        /// Returns the given parent of a <see cref="IDXGIObject"/>, or null if no such object exists.
+        /// </summary>
+        public static T GetParentOrNull<T>(this IDXGIObject dxgiObject) where T : ComObject
+        {
+            // Importantly, this uses the overload with an out value, which itself returns the object.
+            return dxgiObject.GetParent(out T obj).Success ? obj : null;
+        }
```
frenzibyte, line 16: For what it's worth, on iOS, this method results in the AOT toolchain to crash at compilation time. Inlining to usages fixes that.

smoogipoo: Ah yes, thanks for the reminder. Totally forgot about that.

PR #485, opened 2023-03-24
Allow D3D devices to tear if requested

smoogipoo:
Perhaps to be seen as required for https://github.com/mellinoe/veldrid/pull/484 since Microsoft [recommend this while using the flip model](https://github.com/MicrosoftDocs/win32/blob/docs/desktop-src/direct3ddxgi/for-best-performance--use-dxgi-flip-model.md).

This enables `DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING` on the swapchain and allows presenting this way via `GraphicsDevice.AllowTearing`. Doing so forces the sync interval to 0 as required by the documentation [here](https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/dxgi-present).

### DIFF
--- a/src/Veldrid/D3D11/D3D11GraphicsDevice.cs
+++ b/src/Veldrid/D3D11/D3D11GraphicsDevice.cs
@@ -235,7 +235,17 @@ namespace Veldrid.D3D11
             lock (_immediateContextLock)
             {
                 D3D11Swapchain d3d11SC = Util.AssertSubtype<Swapchain, D3D11Swapchain>(swapchain);
-                d3d11SC.DxgiSwapChain.Present(d3d11SC.SyncInterval, PresentFlags.None);
+
+                int syncInterval = d3d11SC.SyncInterval;
+                PresentFlags flags = PresentFlags.None;
+
+                if (_mainSwapchain.TearingAllowed && AllowTearing)
+                {
+                    syncInterval = 0;
+                    flags |= PresentFlags.AllowTearing;
+                }
+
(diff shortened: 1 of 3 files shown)


smoogipoo:
I've expanded on this PR a bit further because I wanted to implement waitable swapchains which changes the flow a bit (recreating the swapchain).

smoogipoo:
> It's probably feasible to just straight up support different PresentMode options across backends

Is the suggestion here to have tearing always turned on because Veldrid can do whatever it wants? I have two concerns here:
1. I think fullscreen tearing is a very specific use-case. A few games I've tested (WoWClassic/Apex Legends/Diablo4) appear to have this turned off. Having it be turned on in general would be fine and match MS's recommendations but:
2. It seems like if you don't use a waitable object for a period of time that the waitable object stops working. I don't know what causes this but it's why I've made the tearing flag recreate the swapchain. For us it meant that switching from fs+tearing to windowed meant that we could no longer benefit from the waitable object.

I'd need to dig deeper to see what causes (2), but if my interpretation of your suggestion is accurate then I'm willing to see if I can get anywhere.

> The right way to expose this part is through signaling/waiting on a Fence

Sounds good. How would you expose the fence? Another method like `CreateNextFrameFence()`?

> should be exposed through the Swapchain object itself

Agree.

smoogipoo:
We're going to continue working on this to our/our users' expectations, resolving issues as we go, and keep this PR updated as things are changed around to fit that.

PR #486, opened 2023-03-24
Fix Vulkan _scImages array not resizing when scImageCount changes

Natelytle:
After initializing the array for _scImages, SetNewSwapchain only checks that _scImages == null. However, after _scImages is set to the length of scImageCount, scImageCount can increase in value, causing an index outside of the bounds of the array error. This change initializes _scImages as an empty array, and the check for null is replaced by a check for if _scImages.Length < scImageCount.

### DIFF
--- a/src/Veldrid/Vk/VkSwapchainFramebuffer.cs
+++ b/src/Veldrid/Vk/VkSwapchainFramebuffer.cs
@@ -16,7 +16,7 @@ namespace Veldrid.Vk
         private uint _currentImageIndex;
 
         private VkFramebuffer[] _scFramebuffers;
-        private VkImage[] _scImages;
+        private VkImage[] _scImages = {};
         private VkFormat _scImageFormat;
         private VkExtent2D _scExtent;
         private FramebufferAttachment[][] _scColorTextures;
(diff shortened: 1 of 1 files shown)


zaafar (approved):

smoogipoo:
I'm still getting a GPU reset when going from unlimited to vsync with this:
```
[40346.633996] [drm:amdgpu_job_timedout [amdgpu]] *ERROR* ring gfx_0.0.0 timeout, signaled seq=5192369, emitted seq=5192371
[40346.634424] [drm:amdgpu_job_timedout [amdgpu]] *ERROR* Process information: process osu.Framework.T pid 584335 thread osu.Framework.T pid 584335
[40346.634809] amdgpu 0000:0b:00.0: amdgpu: GPU reset begin!
[40350.634956] amdgpu 0000:0b:00.0: amdgpu: failed to suspend display audio
[40351.059964] amdgpu 0000:0b:00.0: [drm:amdgpu_ring_test_helper [amdgpu]] *ERROR* ring kiq_2.1.0 test failed (-110)
[40351.060173] [drm:gfx_v10_0_hw_fini [amdgpu]] *ERROR* KGQ disable failed
[40351.285981] [drm:gfx_v10_0_hw_fini [amdgpu]] *ERROR* failed to halt cp gfx
[40351.293578] [drm] free PSP TMR buffer
```
Are you able to repro or believe this is unrelated?

Natelytle:
That is unrelated, this fixes high framerates

PR #489, opened 2023-03-28
Target Net6

Hitsounds:
Update project to target .Net6


zaafar:
the library already target both netstandard2.0 and net6.0, not sure what are you trying to solve over here.

Hitsounds:
Sorry for the frivolous commit. It consists of changes I needed to make to make all the subprojects build but is is quite messy.

PR #494, opened 2023-04-01
Fix D3D11 prematurely disposing swapchain framebuffer texture

frenzibyte:
This is split from https://github.com/mellinoe/veldrid/pull/363 to revive the PR while reducing the scope of the changes, and also includes a fix for the original changes as not disposing the texture at all results in a fatal error when calling `ResizeBuffers` on the next swapchain resize call.

### DIFF
--- a/src/Veldrid/D3D11/D3D11Swapchain.cs
+++ b/src/Veldrid/D3D11/D3D11Swapchain.cs
@@ -145,6 +145,8 @@ namespace Veldrid.D3D11
             Resize(description.Width, description.Height);
         }
 
+        private ID3D11Texture2D backBufferTexture;
+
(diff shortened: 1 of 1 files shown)


bdach:
Mental note for posterity: if this is applied on top of https://github.com/mellinoe/veldrid/pull/485, then the two changes will not combine correctly despite not having merge conflicts. The undisposed swapchain framebuffer texture must be disposed before the swapchain is recreated to enable/disable tearing - otherwise mutating the `AllowTearing` flag will result in a crash due to an access violation.

Just figure it's better to mention it, since we're using both patches in [our fork](https://github.com/ppy/veldrid) and just stumbled on this ourselves.

PR #495, opened 2023-04-06
Added lock to prevent race condition

aquilinum0:
`VkCommandList.CommandBufferCompleted(VkCommandBuffer completedCB)` competes with `VkCommandList.End()` resulting in a race condition:

https://github.com/mellinoe/veldrid/blob/367f3ec52f92ffe65681cafd16cb65d9f89e0af0/src/Veldrid/Vk/VkCommandList.cs#L114
https://github.com/mellinoe/veldrid/blob/367f3ec52f92ffe65681cafd16cb65d9f89e0af0/src/Veldrid/Vk/VkCommandList.cs#L464

### DIFF
--- a/src/Veldrid/Vk/VkCommandList.cs
+++ b/src/Veldrid/Vk/VkCommandList.cs
@@ -461,7 +461,10 @@ namespace Veldrid.Vk
             }
 
             vkEndCommandBuffer(_cb);
-            _submittedCommandBuffers.Add(_cb);
+            lock (_commandBufferListLock)
+            {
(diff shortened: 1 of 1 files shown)


aquilinum0:
For anyone searching/interested the bug manifests as a `System.ArgumentException: 'An item with the same key has already been added. Key: Vulkan.VkCommandBuffer'` on line: https://github.com/mellinoe/veldrid/blob/367f3ec52f92ffe65681cafd16cb65d9f89e0af0/src/Veldrid/Vk/VkCommandList.cs#L107

PR #499, opened 2023-04-14
Avoid encoding redundant Metal render/compute commands to reduce CPU time

frenzibyte:
Xcode has been yelling about this for a while now but I haven't realised how bad the overhead is until I've profiled it myself. This PR reduces down all encoding commands as much as possible to save CPU time.

| Before | After |
|--------|-------|
|  |  |

(late texture creation warning is irrelevant)

### DIFF
--- a/src/Veldrid/MTL/MTLCommandList.cs
+++ b/src/Veldrid/MTL/MTLCommandList.cs
@@ -845,13 +845,23 @@ namespace Veldrid.MTL
             }
         }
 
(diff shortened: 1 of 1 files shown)


frenzibyte:
Also updated `setScissorRects` accordingly, reducing CPU overhead.

Before:

CleanShot 2023-04-15 at 13 37 51

After:

CleanShot 2023-04-15 at 13 38 00

PR #519, opened 2024-01-19
Fix some incorrect line endings

smoogipoo:
For some reason these weren't fixed when `.gitattributes` was added, so these changes now pollute the local work tree.


smoogipoo:
Merging because this is just line endings, and I have other changes in the pipeline (rebasing is silly atm).

PR #522, opened 2024-01-27
Flush vertex bindings when vertex strides change

smoogipoo:
I have a VBO with two different vertex strides inside it.

`SetVertexBufferCore` normally sets `_vertexBindingsChanged = true`, but this isn't happening because the VBO isn't getting rebound.


frenzibyte:
Looks like there are relevant test failures here.

peppy (approved):
I think this looks correct

Issue #117, opened 2018-09-05
[Metal] CommandList.UpdateBuffer with a byte size of 0 causes a crash


geoeo:
`void CommandList.UpdateBuffer<VertexPositionColor>(DeviceBuffer buffer, uint bufferOffsetInBytes, ref VertexPositionColor source, uint sizeInBytes)`

This operation causes a crash with `sizeInBytes=0` on Metal but not in OpenGL.
Also, the GraphicsDevice variant seems to work. At least it doesnt crash the application.

Sadly I dont get a debug trace. 
The apple crash report has this trace

```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   com.apple.driver.AppleIntelKBLGraphicsMTLDriver	0x00007fff2d804778 CHAL_INTEL::ChalContext::ChalExecuteCopyBufferData(MTLIGAccelBuffer*, unsigned int, MTLIGAccelBuffer*, unsigned int, unsigned int) + 84
1   com.apple.driver.AppleIntelKBLGraphicsMTLDriver	0x00007fff2d804704 IGAccelBlitCommandEncoder::copyBufferToBuffer(MTLIGAccelBuffer*, unsigned int, MTLIGAccelBuffer*, unsigned int, unsigned int) + 28
2   com.apple.driver.AppleIntelKBLGraphicsMTLDriver	0x00007fff2d8045f8 -[MTLIGAccelBlitCommandEncoder copyFromBuffer:sourceOffset:toBuffer:destinationOffset:size:] + 288
```

mellinoe:
Thanks for the info. This should be fixed on master now -- let me know if you're still seeing it after updating.

geoeo:
Hey, copyBuffer is fixed but not updateBuffer :sweat_smile:

mellinoe:
@geoeo Hmm, you're still getting a crash? From what I saw, CommandList.UpdateBuffer was only crashing because it internally used CopyBuffer. I'll take a look later today if there's still an issue.

geoeo:
yeah afraid so. Exact same error.

mellinoe:
Interesting. I did try that initially, but it didn't crash on my machine. When I looked at your stack trace, it indicated `copyBufferToBuffer` was causing the problem, which I know will be called in a CommandList.UpdateBuffer call, which DOES crash on my machine. In any case, I think the fix is still to just ignore calls to GraphicsDevice.UpdateBuffer if the size is 0.

[...]

PR #524, opened 2024-02-21
Fix UpdateBuffer not checking zero size

Helco:
Calling `GraphicsDevice.UpdateBuffer` with `sizeInBytes == 0` on Vulkan with a recent validation layer (e.g. SDK 1.3.275.0) causes the validation error `VUID-VkBufferCopy-size-01988`. 

The `CommandList.UpdateBuffer` methods have this check already in place so I guess this was a small oversight.

### DIFF
--- a/src/Veldrid/GraphicsDevice.cs
+++ b/src/Veldrid/GraphicsDevice.cs
@@ -718,6 +718,10 @@ namespace Veldrid
                 throw new VeldridException(
                     $"The data size given to UpdateBuffer is too large. The given buffer can only hold {buffer.SizeInBytes} total bytes. The requested update would require {bufferOffsetInBytes + sizeInBytes} bytes.");
             }
+            if (sizeInBytes == 0)
+            {
+                return;
+            }
             UpdateBufferCore(buffer, bufferOffsetInBytes, source, sizeInBytes);
         }
 


src/Veldrid/GraphicsDevice.cs
```diff
@@ -718,6 +718,10 @@ public void Unmap(MappableResource resource, uint subresource)
                 throw new VeldridException(
                     $"The data size given to UpdateBuffer is too large. The given buffer can only hold {buffer.SizeInBytes} total bytes. The requested update would require {bufferOffsetInBytes + sizeInBytes} bytes.");
             }
+            if (sizeInBytes == 0)
+            {
+                return;
```
frenzibyte, line 723: If one graphics device already fails (vulkan) when performing an "update buffer" call with zero bytes, then this should probably throw a `VeldridException` instead of just silently returning. Especially since this is at the surface of the API.

Helco: I would disagree here. This check is already implemented for the other rather similar APIs like `CommandList.UpdateBuffer`, `CopyBuffer` etc. and was discussed in #117 

For the user ignoring the call for `sizeInBytes==0` is not an error, all data supposed to be updated *was* updated, just that all data is no data in this case.

smoogipoo: I agree with it not throwing an exception. I don't think a device failing is a good reason for _all_ devices to fail, as Veldrid is itself opinionated on how operations are performed so that they conform to the requirements set by the device (the D3D buffer update process as an example).

smoogipoo (approved):

Issue #529, opened 2024-03-29
Vulkan memory barrier around uniform buffer update not covering uniform writes


opticfluorine:
While using the Vulkan backend for Veldrid 4.9.0, I am seeing an out-of-order memory access issue involving uniform buffers on some hardware.

I am submitting a CommandList with multiple Draw calls interspersed with updates to a uniform buffer bound to the vertex shader (the uniform buffer contains texture coordinates into a texture atlas). The uniform buffer was created with `BufferUsage.Dynamic | BufferUsage.UniformBuffer`. On certain setups (Radeon RX 580, Arch Linux) the vertex shader is executed out of order with the update to the uniform buffer. On other setups (GTX 1050, Arch Linux) the correct behavior is observed.

Inspecting the events via RenderDoc, it seems Veldrid is inserting a `vkCmdPipelineBarrier` command between the `vkCmdCopyBuffer` and `vkCmdBeginRenderPass`/`vkCmdDrawIndexed` as expected. This barrier has the following properties:

Property | Value
------------- | ------------
`srcStageMask` | `VK_PIPELINE_STAGE_TRANSFER_BIT`
`destStageMask` | `VK_PIPELINE_STAGE_VERTEX_INPUT_BIT`
`pMemoryBarriers[0].srcAccessMask` | `VK_ACCESS_TRANSFER_WRITE_BIT`
`pMemoryBarriers[0].dstAccessMask` | `VK_ACCESS_VERTEX_ATTRIBUTE_READ_BIT`

Should the `pMemoryBarriers[1].dstAccessMask` also include `VK_ACCESS_UNIFORM_READ_BIT` when the resource being updated is a uniform buffer? It seems to me like this barrier is assuming that the driver/hardware will access the vertex buffer before accessing the uniform buffer, which is perhaps not the case with my setup.

As an aside, thanks for your hard work on Veldrid, I've really been enjoying using it.

opticfluorine:
I forked the `master` branch and modified `VkCommandList::CopyBufferCore` to add the `VK_ACCESS_UNIFORM_READ_BIT`. This change resolved the issue I was seeing.

Would it be helpful for me to submit a pull request? I could also go further and try to limit adding this flag to buffer copies involving uniform buffers, however the `CopyBufferCore` method today does not have this information at hand.

PR #530, opened 2024-03-29
Add UniformRead to Vulkan memory barriers for uniform buffers

opticfluorine:
Fixes issue #529.

The memory barrier inserted after a uniform buffer copy in the Vulkan backend did not ensure proper ordering of uniform buffer reads in the shaders. This pull request adds the appropriate flag when the destination buffer is a uniform buffer.


smoogipoo (approved):
Lgtm, thanks!

